PR #1, opened 2022-01-20
Project Setup

meisekimiu:
This PR sets up our repository. It is heavily based off of [the sftp-service project setup,](https://github.com/PermanentOrg/sftp-service/pull/2) though with the following changes:
- Babel is not used in this project. Since this project is a library and isn't an actual program that runs, we can just use `tsc` to build the project and generate type definitions.
- Because we're not using Babel, Jest is configured with `ts-jest` to run our tests.
- The target node version for this project is 14 and up, since we decided we wanted to target wider compatibility with the original node-sdk. I didn't include 12, since by the time this project has an actual stable release, node 12.x probably won't be supported anymore (and I think we wanted to phase it out soon in `node-sdk` anyway).
- TSConfig is based on Node 14 config, but also has configuration to emit type declaration files as part of the build process (since this is a TypeScript library and all). I also added config to ignore `*.spec.ts` files as part of the build process.

As part of the setup process, I quickly wrote some code in the `src/index.ts` file and tested running an `src/index.spec.ts` file as well in Jest to make sure everything works and the TypeScript compiler is configured properly. Obviously I didn't commit this placeholder code but if you'd like to test some of the choices made in this setup, you can do something similar.

I haven't written an `npm prepare` script, but that will be part of next steps for publishing the SDK to NPM.

One thing I haven't set up that's up for discussion and doesn't need to be added in this initial PR: do we want to set up some kind of documentation tool? I don't know much about those tools, but I assume we should pick one sooner rather than later into the development process.


package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "permanent-sdk",
+  "version": "0.0.0",
+  "description": "Functional Node.js SDK for Permanent.org",
+  "main": "src/index.ts",
+  "scripts": {
+    "lint": "eslint ./src --ext .ts,.tsx",
```
jasonaowen, line 7: We're probably not going to write any React in this repo!
```suggestion
    "lint": "eslint ./src --ext .ts",
```

package.json
```diff
@@ -0,0 +1,46 @@
...
+  },
+  "author": {
+    "name": "Permanent.org",
+    "email": "engineers@permanent.org",
+    "url": "https://www.permanent.org/"
+  },
+  "contributors": [
+    {
+      "name": "Natalie Martin"
+    }
```
jasonaowen, line 23: No email or website?

jasonaowen (approved):
This looks good to me! I was able to check out the branch, build it, lint it, and test it. Nicely, @meisekimiu!

PR #8, opened 2022-02-08
Specify NodeJS version for development

### LINKED ISSUE #7: Node versions
What versions of node do we want this SDK to support (e.g. what should we test against)

What version of node do we want development to occur on?

I'd recommend v16 for development, and v14 and v16 for testing.

For context, the [handy release chart](https://nodejs.org/en/about/releases/)

image

slifty:
This PR specifies that development should happen against NodeJS v16.x.  It also updates the `package-lock` to use the latest format.

The package-lock change occurred when I invoked `npm install` running npm `8.3.0`

Resolves #7


meisekimiu (approved):
These changes look good to me! Is there any reason not to use both `.node-version` and `.nvmrc`? I guess it might clutter things up more. I'm just biased because I use `nvm`.

I'm fine with upgrading the package-lock format, although I remember there being some reason why we didn't do that in some of Permanent's other node projects. As long as we stick with node 16 being the development node version, I think it should be fine?

slifty:
@meisekimiu I would recommend *not* being redundant, since it increases risk of desync down the line, but happy to swap it.

FWIW I use nvm too but I have `.nvmrc` in my global .gitignore because I consider it a personal stack tool (rather than a project tool) -- I then run `ln -s .node-version .nvmrc` so that I get the nvm benefit without that local choice impacting the repo.

I've taken to `.node-version` in the repository itself because it the name is meaningful without knowing about any particular tool.

Let me know what you'd prefer in the end (I know I sound strongly opinionated but I don't mean to impose those opinions on this repo if the above isn't convincing!)

slifty:
I wonder if the other projects were because they're deployed -- and so they have to match whatever version of node lives in deployment land?  I don't wanna break anything, but I do agree if our dev is v16 this updated format should be fine (it may even be necessary!)

meisekimiu:
That's a good explanation, thanks for that! What are your thoughts on putting `.nvmrc` in the local `.gitignore` for permanent-sdk?

slifty:
My *purist* answer is that we shouldn't put local dev tool references in a project `.gitignore`

Or put another way: it's always the individual dev's responsibility to making sure that their personal tool choices don't impact a repo (including the `.gitignore`).  The pragmatic reason for the purist answer is that it's not scalable to ignore every possible local dev env set in `.gitignore` (also for the dev: they should have a setup that lets them conveniently contribute to any project even if that project doesn't ignore `.nvmrc` or `.DS_Store` for instance).

That said, I appreciate how that's going to be a pain since you (and probably other perm devs) aren't ignoring `.nvmrc` globally -- I'm totally fine with adding it for VIP quality of life!  Just say the word.

jasonaowen:
+1 to not supporting 12.x, since it's not long for this world.

> I wonder if the other projects needed the old package format were because they're actually deployed (rather than built and published) -- that might mean they have to match whatever version of node / lock format is expected in prod? I don't wanna break anything, but I do agree if our dev is v16 this updated format should be fine (it may even be necessary!)

The practice and policy I've advocated for in the past is to make sure that the version of npm we use be consistent with the version of Node we use; npm 7 with the new `package-lock.json` format is only shipped with Node 16, and Node 14 uses npm 6 and the old format. I think (correct me if I'm wrong?) that the only time or place this matters is for developers; I don't think using the new format would negatively affect users consuming the SDK from the npm package index via npm 6.

For more historical context, Node 16 is less than a year old, and the other repos predate that. Further, packaged versions always lag upstream availability, and I at least was using the nodesource-packaged Node, which I had pinned to Node 14.x and npm 6.x.

So yes, also +1 to using 16 in dev and using the new format for `package-lock.json`.

> My *purist* answer is that we shouldn't put local dev tool references in a project `.gitignore`

+1 to that general policy - for example, I have my global `.gitignore` configured to ignore vim swap files, and that configuration does not belong in individual repo configurations.

`.nvmrc` is less obviously subject to that policy, but I agree with the plan of "here's why we haven't yet, but if it becomes much more convenient to include rather than exclude then yes let's"

slifty:
I'll merge this in for now since it's approved -- but we can absolutely swap in `.nvmrc` (or add nvmrc to `.gitignore`) any time!

PR #9, opened 2022-02-08
Initial code structure

### LINKED ISSUE #6: Determining code organization
Time to break ground on this repository -- there's no one true way to organize code in an SDK repository, but I wanted to document a proposed structure to start with.

What do we think of something like this:

```
- docs
- src
| - functions
| | - getAccounts.ts
| | - getAccount.ts
| | ...
| | - index.ts // aggregates all the functions so we can do things like "import * from './functions'"
| - util.ts
| - types.ts
| - index.ts // project root, exports the full contents of /api and a subset of /types
```

Under this structure our types would be defined in one place and each individual "function" of the SDK would have its own file.

We would then export anythign public facing from the root level `index.ts`

`functions` may not be the right directory name; some alternatives: `api`, `sdk`, or `lib`.

slifty:
This PR adds an initial skeleton structure to the repository

The intent is that SDK functionality will exist in the sdk directory, and supporting logic will be put into utils.ts

Resolves #6

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,19 @@
 # permanent-sdk
 This project is a TypeScript SDK for interfacing with Permanent.org's API.
 
+## Code
+
+The project is organized as follows:
+
+```
+- docs // Contains documentation related to the project
+- src  // Contains all project code
+| - sdk   // The core / exposed functionality of the SDK
+| - utils.ts
+| - types.ts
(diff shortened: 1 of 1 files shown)


slifty:
Hearing no arguments against I'm gonna just push this forward!

Issue #4, opened 2022-02-07
Implement archive access


slifty:
Once we have an authenticated session we need to be able to `getArchive` and `getArchives` for a given user.

I believe this will involve the definition of a few TypeScript types:

* `ArchiveVO` -- this is the VO returned by the PHP api, for internal use within the SDK.
* `Archive` -- this is the exposed "archive" type that the SDK will return (and accept) for various archive methods.

as well as the following exposed SDK methods:

* `getArchives` - returns a list of populated archives (takes token)
* `getArchive` - returns a populated archive (takes token + archive ID)

A little more context can be found in [our SDK thoughts notes](https://pad.opentechstrategies.com/p/permanent-sdk-thoughts)

slifty:
Let's talk about the archive object!

Here are the attributes that the [PHP ArchiveVO contains](https://github.com/PermanentOrg/back-end/blob/be63d67417b6fe6ed57510e882e48ea6bca9a225/library/base/vo/base.archive.vo.php#L14-L29):

* archiveId
* publicDT
* archiveNbr
* public
* view
* viewProperty
* thumbArchiveNbr
* imageRatio
* type
* thumbStatus
* thumbURL200
* thumbURL500
* thumbURL1000
* thumbURL2000
* thumbDT
* tatus

I'm going to try to figure out what each of these actually is and determine better types / structure and where / how they're used.  The questions I'm exploring right now:

1. What is the "public date time" used for (`publicDT`)?
2. What's the difference between an "archive ID" and an "archive number" (`archiveNbr`)?
3. What is `public`?
4. What is `view` and `viewProperty`?
5. What is `thumbArchiveNbr`?
6. What does `type` refer to and what are the possible values?
7. What are the possible values for `thumbStatus`?
8. What is a `thumbDT` and what is it used for?
9. What are the possible values for `status`?

slifty:
The various type values / possible values are defined in the [constants definition](https://github.com/PermanentOrg/back-end/blob/main/library/definition/perm.constants.php).

Specifically:

> What does type refer to and what are the possible values?

https://github.com/PermanentOrg/back-end/blob/be63d67417b6fe6ed57510e882e48ea6bca9a225/library/definition/perm.constants.php#L411-L414

> What are the possible values for status?

https://github.com/PermanentOrg/back-end/blob/be63d67417b6fe6ed57510e882e48ea6bca9a225/library/definition/perm.constants.php#L320-L328

[...]

PR #10, opened 2022-02-08
Add archive types

slifty:
This PR adds type definitions for the PHP API's `ArchiveVO` as well as the stk API's `Archive` (and related) types.

Some open questions to consider when reviewing:

1. What is the difference between `archiveID` and `archiveNbr` <-- I'd love to NOT expose the SDK users to redundant IDs and instead just use one, but if `archiveNbr` has some kind of meaning that isn't just a redundant ID we should include it (with a better name that reflects it's special meaning)

2. Do we want to put all of our types in a single `types.ts` file, or do we want to break types into their own domain specific files (e.g. `types/archive.ts`) with a `types/index.ts` that flattens the exports.

**NOTE: this includes some things (e.g. the README documentation) from #9 -- please ignore that and just focus on the index.ts + types.ts**


src/types.ts
```diff
@@ -0,0 +1,71 @@
...
+// See:
+// https://github.com/PermanentOrg/back-end/blob/main/library/definition/perm.constants.php
+export enum ArchiveStatus {
+  Deleted = 'status.generic.deleted',
+  Generating = 'status.archive.gen_avatar',
+  Ok = 'status.generic.ok',
+}
+
+export interface ThumbUrl {
+  resolution: 200 | 500 | 1000 | 2000;
```
jasonaowen, line 51: I think it might be more effective to simply call this a `number`, rather than enumerate the specific resolutions that the back-end happens to return right now. I'm not saying it's likely that that list is going to change in the near future, but I don't think it's particularly useful here to make that potential change more difficult. Clients who use this are going to have to iterate through the list of thumbnails to find the resolution that best fits either way.

src/types.ts
```diff
@@ -0,0 +1,71 @@
...
+  url: string;
+}
+
+export interface Thumbnail {
+  urls: ThumbUrl[];
+  readonly status: ThumbStatus;
+  readonly createdAt: Date;
```
jasonaowen, line 58: At some point we'll need to come up with a policy for representing the difference between the models we wish we had and the models we actually have, and that's best done early, I think.

A thumbnail has a particular resolution, URL, status, and creation date. The back-end might conflate some of those, and omit others, but I think we should represent the model we want and deal with the impedance mismatch in our own code. It doesn't make sense to model a "single" `Thumbnail` that has multiple URLs for multiple resolutions; those are separate thumbnails, and they were generated at separate times - even if those times were actually microseconds apart, and even if the back-end didn't record each of those times.

So instead, I would propose:

```typescript
export interface Thumbnail {
  resolution: number;
  url: string;
  status: ThumbStatus;
  createdAt: Date;
}

export interface Archive {
  thumbnails: Thumbnail[];
  //...
}

File: src/types.ts [hunk 3]
```diff
@@ -0,0 +1,71 @@
...
+export interface Archive {
+  id: number;
+  displayDate: Date;
+  type: ArchiveType;
+  thumbnail: Thumbnail;
+  isPublic: boolean;
+  readonly status: ArchiveStatus;
+  readonly createdAt: Date;
+  readonly updatedAt: Date;
+  readonly thumbCreatedAt: Date;
```
jasonaowen, line 70: Is `thumbCreatedAt` the same as `thumbnail.createdAt`? If so, what's the benefit to having it here? If not, what's the difference?

src/types.ts
```diff
@@ -0,0 +1,71 @@
...
+}
+
+export interface Archive {
+  id: number;
+  displayDate: Date;
+  type: ArchiveType;
+  thumbnail: Thumbnail;
+  isPublic: boolean;
+  readonly status: ArchiveStatus;
+  readonly createdAt: Date;
```
jasonaowen, line 68: +1 to the `verbedAt` pattern for dates - I like that!

src/types.ts
```diff
@@ -0,0 +1,71 @@
+interface BaseVO {
+  createdDT: string;
+  updateDT: string;
+  status: string;
+  type: string;
+}
+
+export interface ArchiveVO extends BaseVO {
```
jasonaowen, line 8: Two thoughts: I feel like we should hide the VOs away some how, or if we have to expose them at least figure out a way to indicate that they're internal-only. Also, if we put them all in the same file this file is going to become very large!

src/types.ts
```diff
@@ -0,0 +1,71 @@
...
+  thumbDT: string;
+}
+
+// See:
+// https://github.com/PermanentOrg/back-end/blob/main/library/definition/perm.constants.php
+export enum ArchiveType {
+  Family = 'type.archive.family',
+  Nonprofit = 'type.archive.nonprofit',
+  Organization = 'type.archive.organization',
+  Person = 'type.archive.person',
```
jasonaowen, line 32: I'm afraid I'm actually not super familiar with enums in TypeScript! Can you show me what an example of client code interacting with an archive's type or status would look like? Particularly in JavaScript?

In general, the prefixing scheme the back-end uses for constants feels like something that ought to be hidden from the user; clients ought not have to ever see the substrings `type.archive.` or `status.generic.`. We should translate that for the user, which could then define as a simple `string` union:

```typescript
export type ArchiveType = 'family' | 'nonprofit' | 'organization' | 'person';
```

jasonaowen (commented):
Thank you so much for opening this PR, @slifty! It lets us have a useful conversation around specifics, instead of generalities.

jasonaowen:
> if `archiveNbr` has some kind of meaning that isn't just a redundant ID we should include it (with a better name that reflects it's special meaning)

As @cecilia-donnelly points out in https://github.com/PermanentOrg/permanent-sdk/issues/4#issuecomment-1033128157, `archiveNbr` is user-facing in the form of URLs. 100% agree that we should not expose it under that name, and I love your idea to better reflect its special meaning.

I think there are, broadly, two use cases for which it matters: generating URLs, and parsing URLs. (@meisekimiu, please double-check this assertion!) The former is pretty simple, I think: add a `url` or so to `Archive` that is base URL + `archiveNbr` or whatever. The latter is harder, and I don't remember how web-app handles it.

src/types.ts
slifty: My thinking for having an explicit list is that I'm going to be implementing a translation between the Archive and ArchiveVO type -- and at that point I am going to map ONLY thumbs with one of these four resolutions (there's no place for other resolutions).  In that sense, there are no other resolutions and so calling it a full blown number might be a bit misleading / lead to bugs.

This would also allow us to communicate changes to SDK users in future -- if we remove a resolution they would be alerted by tooling (and if we add / expand to numbers it would be backwards compatible).

Thoughts?

src/types.ts
slifty: Great change!

src/types.ts
slifty: Oh I meant to take this out (this was in my first version before I started moving to the alluded to "models we wish we had" approach)

src/types.ts
slifty: This won't be exposed via SDK (it's not exported in index.ts) but we do need it for our internal code -- it's what is provided by the PHP api so I thought we would it as written so that I can write a `createArchiveFromArchiveVo` method, for instance.

src/types.ts
slifty: We're going to need to translate to and from those strings at some point in the SDK since that's what the server / API will be returning, so they need to exist somewhere.

I had imagined we would create some constants that mirror these enums, but if we can avoid that constant entirely by just using the strings, and then mapping to / from those types in the VO Type => SDK Type translation method I'm all for it.

(I believe that string union will unfortunately violate our linting rules but could be wrong.)

src/types.ts
jasonaowen: > implementing a translation between the Archive and ArchiveVO type

Thumbnails are read-only, aren't they? That is, the back-end ought not allow changes to this value!

> This would also allow us to communicate changes to SDK users in future -- if we remove a resolution they would be alerted by tooling (and if we add / expand to numbers it would be backwards compatible).

If we wanted to remove a resolution, that would become a breaking change; we would need to coordinate with all users to say "hey, 500px thumbnails are going away in {time period}". But I don't think it should be: the algorithm clients use to select a thumbnail should work equally well if there are 1, 4, or 40 different resolutions available, and they oughtn't care about 500px thumbnails specifically. Right?

src/types.ts
jasonaowen: I should have said this earlier, but: I love the idea of turning this into a list, with the resolution as a property! That's a great way to represent that, and I think it will be much easier for clients to select a thumbnail, as I allude to in the other thread.

src/types.ts
jasonaowen: Oh, yes, sorry - I might be getting confused between user-facing and internal-facing things. We will definitely have to translate at some point, and so will need to internally know that mapping somehow, and this is a fine way to do that. But I don't want to expose that mapping to clients if we can help it!

src/types.ts
jasonaowen: (And I think, off the top of my head, that the algorithm is going to be something like:

```typescript
const getThumbnail = (thumbs: Thumbnail[], targetResolution: number) => {
  if (thumbs.some((thumb) => thumb.resolution > targetResolution)) {
    return thumbs
      .filter((thumb) => thumb.resolution > targetResolution)
      .sort((a, b) => a.resolution > b.resolution)
      [0];
  } else if (thumbs.length > 0) {
    return thumbs.sort((a, b) => a.resolution > b.resolution)[0];
  } else {
    return placeholderThumbnail;
  }
};
```
)

jasonaowen: ...so it would probably be useful to keep our array sorted by resolution, since the order we populate it by is arbitrary.

slifty: OH!  I didn't realize this is read only -- this should totally be a number then (and thumbnails should be marked as read only in the interface, yes?)

(Now I'm imagining a world where TypeScript makes it possible to require ascending order in an array........)

src/types.ts
slifty: You're totally fine -- I think as currently implemented the user would indeed have to know the "internal" string ("know" here could mean "load it via a constant or manually enter the string").

By using something more pithy (like what you suggest) we can avoid use of constants AND avoid the user needing to know that internal string.

Note this assumes we prefer something like:

```
archive.archiveType = 'family'
```

vs

```
archive.archiveType = ARCHIVE_TYPE.FAMILY
```

^ I think we prefer the former to the later, right?

mithuna:
> > if `archiveNbr` has some kind of meaning that isn't just a redundant ID we should include it (with a better name that reflects it's special meaning)
> 
> As @cecilia-donnelly points out in [#4 (comment)](https://github.com/PermanentOrg/permanent-sdk/issues/4#issuecomment-1033128157), `archiveNbr` is user-facing in the form of URLs. 100% agree that we should not expose it under that name, and I love your idea to better reflect its special meaning.
> 
> I think there are, broadly, two use cases for which it matters: generating URLs, and parsing URLs. (@meisekimiu, please double-check this assertion!) The former is pretty simple, I think: add a `url` or so to `Archive` that is base URL + `archiveNbr` or whatever. The latter is harder, and I don't remember how web-app handles it.

ArchiveNbr is a string and not an ID, but it is unique for each archive and its records. Adding [my comment on ArchiveNbr here ](https://github.com/PermanentOrg/permanent-sdk/issues/4#issuecomment-1033875758)

src/types.ts
cecilia-donnelly: The former definitely seems friendlier for users.

meisekimiu (commented):
Is there any reason why we're not including the `fullName` field on this implementation `Archive`/`ArchiveVO` types? In my head the proof of concept code once we implement the HTTP calls would be to just `console.log` out a human readable list of a user's archives, and we can't really do that without `fullName`. <sub>*(I mean, we could just print out archive ID and... thumbnail URLs, but I think having the Archive Name you see in the UI is an important field too!)*</sub>

There are a bunch of other Profile Item fields on the backend `ArchiveVO` class that may or may not be unused (I think they were replaced by `ProfileItemVOs` maybe? I have no idea), but I know `fullName` is used throughout the web-app and shouldn't be ignored. (Though I might rename it for the `Archive` type)

src/types.ts
```diff
@@ -0,0 +1,71 @@
...
+  updateDT: string;
+  status: string;
+  type: string;
+}
+
+export interface ArchiveVO extends BaseVO {
+  archiveId: number;
+  publicDT: string;
+  archiveNbr: number;
+  public: string;
```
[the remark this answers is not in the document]
slifty: Ty ty!

So...  oh goodness.  I used BaseArchiveVO when defining the type... Why... why are there two... what's the difference X(.

Seeing the api version, are a *ton* of fields that I don't have here, and I will update this PR to include them!

> Why is public here a string? Shouldn't it be a [boolean?](https://github.com/PermanentOrg/back-end/blob/be63d67417b6fe6ed57510e882e48ea6bca9a225/api/core/archive/vo/archive.vo.php#L53)

Was a string purely because I didn't know what it was and string was a good catchall (the baseArchiveVO was all I knew about and it said "mixed" for everything so... lol)

As far as exposure I do wanna clarify that the `ArchiveVO` object is an SDK-internal type -- it won't be exposed!  Happy to make it read only to signal to us the intent, but it's most important for `isPublic` to be readonly on the `Archve` (not `ArchiveVO`) type so I'll go add that in.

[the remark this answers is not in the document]
cecilia-donnelly: I'm not sure if you're serious on the "why are there two" but it looks to me like `BaseArchiveVO` is used for the shared fields between `ArchiveVO`, `ArchiveNbrVO`, and `ArchiveMyVO` [I have never seen this last one used - has anyone else?].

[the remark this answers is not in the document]
cecilia-donnelly: Wait...the other two don't even extend BaseArchiveVO so never mind, I have no idea why it exists.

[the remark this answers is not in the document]
slifty: ^ The second set of 👀 should be 
image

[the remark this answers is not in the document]
slifty: I've added all the fields from the full blown VO and updated public to be a bool

src/types.ts
```diff
@@ -0,0 +1,127 @@
...
+
+export interface Thumbnail {
+  readonly resolution: number;
+  readonly url: string;
+  readonly status: ThumbStatus;
+  readonly createdAt: Date;
+}
+
+export interface Archive {
+  id: number;
```
slifty, line 111: This should probably be read-only, right?  Since the ID will either be assigned by server, not exist, or in the case of an object being used for an update would be assigned the one time.

src/types.ts
jasonaowen: Maybe yes! Can you please look into how it would work from a JavaScript rather than TypeScript client? Possibly it would just work, in which case, great!

src/types.ts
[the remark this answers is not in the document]
jasonaowen: > `ArchiveMyVO` [I have never seen this last one used - has anyone else?].

Looks like it was used in the old front end, but neither the current front-end or the mobile apps know about it. Let's delete it!

src/types.ts
jasonaowen: I agree!

However, this raises a concern I hadn't thought through yet, and which isn't obvious during type definitions without accompanying functionality: in the user types, we will need to distinguish between what we accept and what we emit. Or, put more concretely, what does the SDK version of `createArchive` look like? If we accept an `Archive`, then the user cannot meaningfully enter an `id`; if we make it optional, then they will be plagued with null checks when using SDK-created `Archive`s.

How do other SDKs handle this?

slifty: I'll take a look -- but I assume that `createArchive` wouldn't accept an archive, and would instead have its parameters (and types) defined as part of the function API without an explicit named type.

src/types.ts
```diff
@@ -0,0 +1,138 @@
...
+  slug: string;
+  displayDate: Date;
+  type: ArchiveType;
+  name: string;
+  description: string;
+  readonly children: (Folder | Item)[];
+  readonly accessRole: AccessRole;
+  readonly usedSpace: number;
+  readonly availableSpace: number;
+  readonly availableFiles: number;
```
jasonaowen, line 132: Likewise, are these aspects of an archive? I thought that our space accounting was per-account rather than per-archive, although there's some desire to [make funding more sophisticated](https://permanent.atlassian.net/browse/PER-8489).

src/types.ts
jasonaowen: Great, that sounds like a good approach, and so these types are specifically for what we return!

src/types.ts
slifty: Happy to take this out -- I wasn't sure if archives were granted specific quotas.  (I don't think we should define this type in terms of nonexistent functionality; we can always add that if it becomes real)

src/types.ts
```diff
@@ -0,0 +1,138 @@
...
+
+export interface Archive {
+  id: number;
+  slug: string;
+  displayDate: Date;
+  type: ArchiveType;
+  name: string;
+  description: string;
+  readonly children: (Folder | Item)[];
+  readonly accessRole: AccessRole;
```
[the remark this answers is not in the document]
slifty: If it's related to the user's relationship with *ANY* archive then I completely agree it doesn't belong.

If it's related to the user's relationship with *THIS* archive then I think this is where it belongs, but:

We could rename this to `accessRoles` and have it be an array of `AccessRoles` which are defined as having a `userId` and a `role`

That way we're respecting the fact that it's about a relationship -- and if the user is allowed to know ALL access roles that list would be populated with more.

Thoughts?

src/types.ts
```diff
@@ -0,0 +1,138 @@
...
+  company: string;
+  description: string;
+}
+
+/*************************
+ ** Everything ABOVE this comment is internal to the SDK and used to interface with the
+ ** permanent API.
+ **
+ ** Everything BELOW this comment is part of the SDK's external API.
+ **************************/
```
jasonaowen, line 66: Horizontal rules in comments make me sad. :cry:

This file is already over 100 lines, and that's just for the first type. Can we split this file up? I think a bunch of smaller files, split across user-facing and API-facing lines, would make sense; maybe something like:

- `src/types/api/archivevo.ts`
- `src/types/api/itemvo.ts`
- `src/types/sdk/archive.ts`
- `src/types/sdk/folder.ts`
- etc

It leads to a bunch more files, which I think makes for more discoverability; as I mentioned during our call the other day, sometimes I do just start poking around my `node_modules` to learn how something works, and I think I would find those files easier to navigate than the massive file `types.ts` will inevitably turn into.

src/types.ts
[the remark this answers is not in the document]
slifty: Another approach could be to treat `GrantedAccessRole` (or whatever) as a first class object with an SDK function to retrieve the list, and at that point yeah we could either (A) remove it from this object or (B) have some kind of parameter to signal the desire to make that extra call.

We could also just start with the above for *now* and when we implement plan to tweak based on the implementation decisions of that retrieval.  (I raise that path simply to avoid boiling the entire ocean with this PR  -- though if we need this to be in the pot that's fine.)

src/types.ts
jasonaowen: Haha, I appreciate the consideration! Glad we're on the same page.

src/types.ts
slifty: Looks like, unlike types, it builds to an object!

```
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.ThumbnailStatus = void 0;
// These statuses are defined in the permanent backend:
// https://github.com/PermanentOrg/back-end/blob/main/library/definition/perm.constants.php
var ThumbnailStatus;
(function (ThumbnailStatus) {
    ThumbnailStatus["Generating"] = "status.archive.gen_avatar";
    ThumbnailStatus["Ok"] = "status.generic.ok";
})(ThumbnailStatus = exports.ThumbnailStatus || (exports.ThumbnailStatus = {}));
```

(so "just works")

Given that I'll mark this as resolved for now.

src/types/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
+  "rules": {
+		"sort-exports/sort-exports": [2, { "sortDir": "asc" }]
```
slifty, line 3: We can decide if we want this -- I borrowed the pattern from another project where I found it useful.  It basically means our `index.ts` files will have exports in the same order as the folder they're globbing.

src/types/api/ArchiveVo.ts
```diff
@@ -0,0 +1,43 @@
...
+  thumbStatus: string;
+  thumbURL200: string;
+  thumbURL500: string;
+  thumbURL1000: string;
+  thumbURL2000: string;
+  thumbDT: string;
+
+  // These fields map to those defined in the api ArchiveVO
+  // https://github.com/PermanentOrg/back-end/blob/main/api/core/archive/vo/archive.vo.php
+  childFolderVOs: ChildFolderVo[];
```
slifty, line 27: Note for this (and other VOs) I kept the `VO` instead of `Vo` intentionally to make the serialization / deserialization more straightforward.  This also explains the reason `archiveId` and `publicDT` are seemingly inconsistent (it's because that inconsistency is at the API level)

src/types/api/ArchiveVo.ts
```diff
@@ -0,0 +1,43 @@
+import type { BaseVo } from './BaseVo';
+import type { ChildFolderVo } from './ChildFolderVo';
+import type { FolderSizeVo } from './FolderSizeVo';
+import type { ItemVo } from './ItemVo';
+
+export interface ArchiveVo extends BaseVo {
+  // These fields map to those defined in the library ArchiveVO
+  // https://github.com/PermanentOrg/back-end/blob/main/api/core/archive/vo/archive.vo.php
```
cecilia-donnelly, line 8: https://github.com/PermanentOrg/back-end/blob/main/library/base/vo/base.archive.vo.php

src/types/api/ArchiveVo.ts
cecilia-donnelly: I believe the type here is a FolderVO (not a ChildFolder) on the backend, though I haven't gone through everything here yet to check whether that's a meaningful distinction for the SDK.

src/types/sdk/AccessRole.ts
```diff
@@ -0,0 +1,13 @@
+// These roles are defined in the permanent backend:
+// https://github.com/PermanentOrg/back-end/blob/main/library/definition/perm.constants.php
+export enum AccessRole {
+  Administrator = 'access.role.admin',
+  Contributor = 'access.role.contributor',
+  Curator = 'access.role.curator',
+  Editor = 'access.role.editor',
+  Manager = 'access.role.manager',
+  Minor = 'access.role.minor',
```
cecilia-donnelly, line 9: We never use this on the backend - can we skip it here?

src/types/sdk/AccessRole.ts
```diff
@@ -0,0 +1,13 @@
...
+  NonOwner = 'access.role.nonOwner',
```
cecilia-donnelly, line 10: Ditto.

cecilia-donnelly (approved):
Thanks @slifty !  Exciting!

I may be misunderstanding your intention, organizationally, or I might have missed something on the backend, but: I don't believe the api has a specific "ChildFolderVO" type.  These are just FolderVOs.  Does it make sense to call the API type a FolderVO, then, and just express that it's a child on the SDK side?

src/types/sdk/AccessRole.ts
slifty: For sure -- should we / can we also delete it from the PHP to ensure it never sneaks in?  (This question applies to the Minor role as well)

src/types/api/ArchiveVo.ts
slifty: Good catch ty!

src/types/sdk/AccessRole.ts
cecilia-donnelly: https://permanent.atlassian.net/browse/PER-8627 is the ticket for removing this from the backend

src/types/sdk/AccessRole.ts
cecilia-donnelly: https://github.com/PermanentOrg/back-end/pull/31 please double check me but I think that's all there is to it!

src/types/api/ArchiveVo.ts
jasonaowen: I know this is late, but I think a pattern that we should follow, and that this is an example of, is keeping API-facing front-end types named the same as their corresponding types in the API. So definitely +1!

PR #13, opened 2022-02-25
Add Codecov to CI

### LINKED ISSUE #12: Add codecov reports to our pull requests
[Codecov](https://codecov.io) is a tool that generates code coverage reports for PRs.

We recently removed this tool from another repository, but that was because it was broken -- not necessarily because we don't believe in the value of test coverage analysis.

Let's try it out here.  If it's not valuable or otherwise distracting we can remove it.

slifty:
This PR adds [Codecov](https://codecov.io) to this repository

It also refactors the test CI step to separate "regression" tests from the "main" test suite.  This allows us to avoid repeating our coverage report generation for every version of Node we want to test against.

Note: Right now codecov is saying it's "missing base report" because we don't have any report for the `main` branch to compare the PR against.  I believe this will be resolved after merging.

Resolves #12

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,7 +10,8 @@ jobs:
         with:
           node-version: 16.x
       - run: npm install --production=false
-      - run: npm test
+      - run: npm test -- --coverage
+      - uses: codecov/codecov-action@v2
 
   run_regression_tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,5 +1,10 @@
 name: test
-on: [pull_request, workflow_dispatch]
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
 jobs:
   run_tests:
     runs-on: ubuntu-latest
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -3,9 +3,20 @@ on: [pull_request, workflow_dispatch]
 jobs:
   run_tests:
     runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        name: Test Node.js 16.x
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 16.x
+      - run: npm install --production=false
+      - run: npm test
+
+  run_regression_tests:
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [14.x]
     steps:
       - uses: actions/checkout@v2
         name: Test Node.js ${{ matrix.node-version }}
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,5 +1,10 @@
 name: lint
-on: [pull_request, workflow_dispatch]
+on:
+  pull_request:
+  workflow_dispatch:
+  push:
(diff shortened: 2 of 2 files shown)


jasonaowen:
> Separate the "primary" node version tests from the "regression" version tests

I like this option! Right now we only have two versions in the matrix, right? 14.x and 16.x? 12.x is about to EOL so we haven't bothered with it, but 18.x will enter LTS in April, so I suppose that only-two-versions matrix is a limited time thing.

So yes: primary with coverage + regression versions without sounds like an excellent plan!

.github/workflows/test.yml
```diff
@@ -13,4 +13,9 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
       - run: npm install --production=false
-      - run: npm test
+      - run: npm test -- --coverage
+      - name: Upload to Codecov
+        run: |
+          curl -Os https://uploader.codecov.io/latest/linux/codecov
+          chmod +x codecov
+          ./codecov
```
jasonaowen, line 21: Is this really how they recommend running it? This is basically just `curl | sh`! Are there any other options?

slifty: It really is, isn't it.

They do, I swear, but agreed: let's do better.

Just for fun here's their instructions:

image


Also it looks like I misread the "deprecation warning" on their official github action, so instead of inventing it I may just use theirs: https://github.com/marketplace/actions/codecov

"This version will no longer function" was referring to version 1...  phewf.

jasonaowen: "Here're some copy-pastable instructions! Oh but you should be more secure than that, go read this link" :grimacing:

jasonaowen: Oh but the GitHub Action looks great!

slifty: (I had included a codecov token related to a dead example repo in my screenshot / original message -- it wasn't related to permanent, and the repo truly was a throwaway repo, but I've deleted it anyway)

.github/workflows/test.yml
```diff
@@ -2,13 +2,25 @@ name: test
...
+  run_regression_tests:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [14.x]
     steps:
       - uses: actions/checkout@v2
-        name: Test Node.js ${{ matrix.node-version }}
+        name: Test Node.js ${{ matrix.node-version }} (regression)
```
slifty, line 23: Just realized: I'm gonna take out this change -- I think we already make it clear that it's a regression in our job name, and this name only shows up in logs, where it's not particularly useful or worth the extra characters.

jasonaowen:
Sorry, one more question: how does Codecov get reports on the `main` branch? As written, I think this is only running the tests on PRs. Do we also need to run that test on merges to `main`?

slifty:
It's a good question -- I was hoping it would magically know that the branch was merged and allocate, but I could see us wanting to run tests on main as well as good practice regardless.

What do you think of me adding one more patch to the CI to run on main?

jasonaowen:
> What do you think of adding one more patch to this PR to the CI to run tests on main (may be worth running lint too while we're at it)?

I think that would be good!

slifty:
@jasonaowen done!  Ready for re-review for real.

jasonaowen (approved):
This looks good to me! Thanks, @slifty!

Issue #5, opened 2022-02-08
Can we be agnostic about HTTP libraries?


jasonaowen:
If reasonably possible, I think it would be good for this SDK to not explicitly depend on any particular library for making HTTP requests. The old SDK uses [axios](https://www.npmjs.com/package/axios), and that is a fine choice, but it makes it less suitable for use in, say, Angular applications. The [next major version of Node will include support for `fetch`](https://fusebit.io/blog/node-fetch/), but if we want to support current LTS versions of Node, we won't be able to reliably use that until 2024.

Is there a way for us to be agnostic about HTTP libraries? Can we accept some kind of argument somewhere that makes requests and returns results in a known and expected way? If so, is such an approach worth the additional complexity?

jasonaowen:
After 20 minutes of research, I don't see any clear choices; this is not obviously a pattern that exists in the Node ecosystem. I'll leave this open for the rest of the team to consider, but it's quite possible that this idea is not practical.

slifty:
Since we're using node fetch I think I'll close this issue for now -- we can re-visit when we move to Node 18 and want to drop support for Node 16!

PR #14, opened 2022-02-25
Add ArchiveVO to Thumbnails conversion utility

slifty:
This PR adds our first "API to SDK" conversion, which turns an ArchiveVO into an array of Thumbnails

It also includes some necessary patches as well as a few pieces of general tooling needed to support that conversion.

Some key decisions I would like feedback on, as they represent patterns for future code:

1. Defining the factory and conversion tools in a root level `utils` directory.
2. The naming pattern for the factory methods (`createX`)
3. The naming pattern for the conversion methods (`xToY`, as opposed to something like `convertXtoY`)
4. The decision to create a *client side only* `Unknown` value for *each enum*.

Highlighting the "unknown enum" value in particular:

The alternative to the currently written implementation would be for the conversion script to throw an error if it is provided a value it doesn't expect.  I can see arguments for either approach.  The error approach would be more explicit AND would make the type / enum more accurate, but it would also require the developer to know to catch the error in order to avoid a broken user experience.

_Note: This is ready for review, but I would like to rebase it off of #13 once that is merged, so we can (hopefully) see a working test coverage report._


src/utils/__test__/__snapshots__/createArchiveVo.test.ts.snap
```diff
@@ -0,0 +1,79 @@
...
+}
+`;
+
+exports[`createArchiveVo should create an ArchiveVo object with no parameters 1`] = `
+Object {
+  "accessRole": "",
+  "archiveId": 0,
+  "archiveNbr": 0,
+  "birthDay": "",
+  "childFolderVOs": Array [],
```
slifty, line 48: Ideally this and the other VO arrays would have values in this test to confirm they are being written properly... but that would involve first having the factories for each VO type.

When those are added I will try to remember to update these tests.

package.json
```diff
@@ -43,5 +43,8 @@
     "jest": "^27.4.7",
     "ts-jest": "^27.1.3",
     "typescript": "^4.5.4"
+  },
+  "dependencies": {
+    "node-fetch": "^3.2.0"
```
slifty, line 48: This isn't related to the functionality of this PR and snuck in -- I can it out, or if we want to include it now that's fine too; it's going to need to be added for both my and Kavya's work so maybe better to get it merged sooner than later.

src/utils/archiveVoToThumbnails.ts
```diff
@@ -0,0 +1,50 @@
...
+} from '../types';
+
+import { ThumbnailStatus } from '../types';
+import { isThumbnailStatus } from './isThumbnailStatus';
+
+function statusCodeToThumbnailStatus(statusCode: string): ThumbnailStatus {
+  if (isThumbnailStatus(statusCode)) {
+    return statusCode;
+  }
+  return ThumbnailStatus.Unknown;
```
slifty, line 13: This line is currently untested -- I'll wait until we finalize our decision to include `Unknown` (as opposed to throw an error here) before writing a test.

package.json
jasonaowen: Thanks for splitting it into its own commit! I'm fine with leaving it in, but maybe it would be good to link to #5 in the commit message?

src/types/api/ArchiveVo.ts
```diff
@@ -9,7 +9,7 @@ export interface ArchiveVo extends BaseVo {
   archiveId: number;
   publicDT: string;
   archiveNbr: number;
-  public: boolean;
+  isPublic: boolean;
```
jasonaowen, line 12: > Unfortunately "public" is a reserved word in TypeScript which means we can't use it in this way.  This divergence might cause minor confusion down the line but our hands are somewhat tied.

Can we quote it here?

```typescript
  'public': boolean;
```

src/utils/generateEnumTypeguard.ts
```diff
@@ -0,0 +1,5 @@
+// Copyright CC BY-SA 4.0
+// https://creativecommons.org/licenses/by-sa/4.0/
+// https://stackoverflow.com/a/58278753/159522
+export const generateEnumTypeguard = <T>(e: T) => (token: unknown): token is T[keyof T] =>
```
jasonaowen, line 4: Oof, this is immediately very complicated.

First, we need to credit the author; see [Is it legal to copy Stack Overflow questions and answers?](https://meta.stackexchange.com/a/24618). It's good that you're already citing the license and the answer URL!

Second, CC-BY-SA is a copyleft license that would restrict our ability to relicense this repo under something other than the AGPL or GPL, which I believe we want the option to do - it's not obvious to me that we necessarily want to force our partners to release their source code by using our SDK.

Third, is this or something like it available as a package? I would rather not maintain such code if we could avoid it.

jasonaowen (changes requested):
I'm only partway through reviewing this PR, but I'm pretty concerned about the `generateEnumTypeguard`, so wanted to start that conversation sooner rather than later.

src/types/api/ArchiveVo.ts
slifty: Oh hey yeah that'd do the trick too -- it will require us to also quote in places we create new archiveVos (our linter will remind us of this problem every time we forget, though not the solution).  This is internal to the SDK of course, so that's probably OK.

Neither solution is completely perfect -- the benefit of `isPublic` is that the problem footprint is limited: just the one place we're parsing the server response and populating the object, whereas using `'public'` will be necessary any time we create an archiveVo...  That said, `'public'` would match the server API more directly.

I'll give it a whirl (if the above has made you reconsider the suggestion let me know but I imagine you had already considered those things).

src/utils/generateEnumTypeguard.ts
slifty: It *is* ugly and complicated without a doubt.  I bet I could de-code-golf it and it would at least be a bit more clear.  I wonder if doing that would make it so we can claim the copyright........  Or is the idea of a typeguard generator now forever copyleft.

I'll check if there's a package -- I'm doubtful -- though we (or I...) could consider releasing one.

jasonaowen: >  I wonder if doing that would make it so we can claim the copyright........ Or is the idea of a typeguard generator now forever copyleft.

The way a previous company I worked at handled a nearly identical situation was to ask someone who hadn't seen the code to independently re-implement it. If we do that, you and I are tainted by having seen and studied this solution; we'd have to ask someone else (who I will not tag).

src/types/api/ArchiveVo.ts
jasonaowen: Awesome! Thanks for doing the heavy lifting; I wasn't completely sure if that solution would actually work in an interface declaration. If it does work, I think that'd be the best option - I want to keep the API-facing data types as close to the back-end's definitions as possible, right down to the capitalization.

src/utils/generateEnumTypeguard.ts
slifty: If it helps, I've already forgotten the original code; I mean I remember every time my eyes scan up this page but then I forget again moments later.

src/types/api/ArchiveVo.ts
slifty: Actually -- I just realized.

Typescript doesn't care about this definition being a reserved word -- the linter didn't complain at all in the original PR.

The linter is upset when I *create an object* with this interface.  Given that, I'd just ditch this commit completely and update the other lines to use `'public'`.

src/utils/generateEnumTypeguard.ts
jvasile: Hi all, mind if I weigh in?  I know I'm not the reviewer here, but I want to point out some resources that might help.

The SO snippet is under CC By-SA 4.0.  SO has used that license for Subscriber Content (see the section 6 of the TOS at [1]) since 5 Sept 2019 (see [2]).  The SO post and answer are dated 8 Oct 2019 (see [3]).

I question whether this type of one-liner is enough to qualify for copyright protection when used like this, but if you want to be super conservative (and you know how much I love when people are super conservative), consider this path forward that avoids cleanrooming a one-liner:  you can level up CC By-SA toward toward AGPL.  First, note that CC-By-SA 4.0 is compatible with GPLv3 (see [4]).  Then note that one can include GPLv3-licensed code in AGPLv3 works. (see AGPLv3, section 13 at [5]).

Attribution is required, just as with many other licenses.  This commit appears to include that attribution.  It might be extra polite to name the SO user, jcalz, in the attribution comment.  Other than that, do you know any reason Permanent can't treat this like any other third-party code?

[1] https://stackoverflow.com/legal/terms-of-service
[2] https://meta.stackexchange.com/questions/333089/stack-exchange-and-stack-overflow-have-moved-to-cc-by-sa-4-0
[3] https://stackoverflow.com/questions/58278652/generic-enum-type-guard/58278753#58278753
[4] https://wiki.creativecommons.org/wiki/ShareAlike_compatibility:_GPLv3
[5] https://www.gnu.org/licenses/gpl-3.0.en.html

slifty: I'm resolving this conversation since the core issue is addressed

cecilia-donnelly (commented):
Thanks @slifty!  I'm happy with your choices on 1-3 (in your PR message).  The "Unknown" is fine - I'm not sure what we use thumbnail status for, but "unknown" should be as useful as an error to indicate that something went wrong.

src/utils/generateEnumTypeguard.ts
jasonaowen: Sorry to unresolve, but I think this is a sticking point:

> if this SDK is agpl does that mean any code that uses the package must also be agpl?

My understanding is yes, unless we grant users of this package a separate license. Permanent is only able to grant such licenses if they own all the IP, or if the third-party IP is licensed compatibly.

I believe including code licensed under CC-BY-SA 4.0 would prevent that.

I also believe that it is important to Permanent that clients are able to use this SDK from non-[A]GPL code - is that correct, @cecilia-donnelly?

@jvasile, I'd love for you to double-check my reasoning, here!

cecilia-donnelly: @jasonaowen we (@slifty and I) talked to @kfogel in our regular meeting and he thought this was too small a piece of code to be infringing.  Also it looks like @slifty rewrote it in any case so I'm doubly unconcerned.  Obviously also interested in @jvasile's thoughts!

src/utils/generateEnumTypeguard.ts
```diff
@@ -0,0 +1,10 @@
+// This is a modified version of code written by StackOverflow user @jcalz
+// Copyright CC BY-SA 4.0
+// https://creativecommons.org/licenses/by-sa/4.0/
+// https://stackoverflow.com/a/58278753/159522
+export const generateEnumTypeguard = <GenericEnum>(genericEnum: GenericEnum) =>
+  (value: unknown): value is GenericEnum[keyof GenericEnum] => {
+    const token = value as GenericEnum[keyof GenericEnum];
+    const validTokens = Object.values(genericEnum);
+    return validTokens.includes(token);
+  };
```
jasonaowen, line 10: This is certainly a lot more readable than the previous version - thanks for rewriting, @slifty!

src/utils/createArchiveVo.ts
```diff
@@ -0,0 +1,47 @@
...
+  updateDT: '',
+  status: '',
+};
+
+export function createArchiveVo(input: Partial<ArchiveVo> = {}): ArchiveVo {
+  return {
+    ...defaultValues,
+    ...input,
+  };
+}
```
jasonaowen, line 47: I know I initially suggested this approach, but now that I look at it in practice, what would you think about instead exporting the `defaultValues` object and expecting (internal) users to just do the object spread themselves? That'd save us some tests and a function call.

If not, this can be written as a simple arrow function expression:

```typescript
export const createArchiveVo = (input: Partial<ArchiveVo> = {}): ArchiveVo => ({
  ...defaultValues,
  ...input,
});
```

src/utils/archiveVoToThumbnails.ts
```diff
@@ -0,0 +1,50 @@
...
+  if (archiveVo.thumbURL200) {
+    resolutionUrlTuples.push([
+      200,
+      archiveVo.thumbURL200,
+    ]);
+  }
+  if (archiveVo.thumbURL500) {
+    resolutionUrlTuples.push([
+      500,
+      archiveVo.thumbURL200,
```
jasonaowen, line 29: This should be `thumbURL500`, not `200`, right?

src/utils/archiveVoToThumbnails.ts
```diff
@@ -0,0 +1,50 @@
...
+
+function statusCodeToThumbnailStatus(statusCode: string): ThumbnailStatus {
+  if (isThumbnailStatus(statusCode)) {
+    return statusCode;
+  }
+  return ThumbnailStatus.Unknown;
+}
+
+export function archiveVoToThumbnails(archiveVo: ArchiveVo): Thumbnail[] {
+  const resolutionUrlTuples: [number, string][] = [];
```
jasonaowen, line 17: Rather than building an empty array and conditionally pushing to it, I like having an [array literal with conditional elements](https://stackoverflow.com/questions/50280750/define-javascript-array-literal-with-conditional-elements).

With that approach, this entire function could also be a simple arrow function expression!

src/utils/generateEnumTypeguard.ts
kfogel: It's a one-liner that probably is too small to be infringing anyway, and anyway we probably want to de-code-golf it, which would take it further away from the original source (though I suggest that, out of courtesy, we retain credit to the original SO author as inspiration).

So I think we can include it, rearranged/rewritten for clarity, and be quite safe with respect to potential future relicensing of the SDK as a whole.

CC'ing @jvasile here to get his take.  I have written the above mainly to give him a convenient thing to "+1" or thumbs-up if he agrees with it :-).

jasonaowen (commented):
It might be good for us to agree on an overall approach of function definition style!

I have come to prefer arrow functions, and ideally simple arrow function expressions when possible (eg, `(a) => a` or, for returning objects, `(a) => ({ b: a })`). Arrow functions have a less surprising set of rules than classic functions: there is no confusion to be had around `this`, they can't be used as an object constructor, and they don't support the rarely-used special `arguments` object. Simple arrow function expressions additionally communicate a relative simplicity - as their name would suggest - and I have found that trying to use them leads to smaller, clearer functions than I might otherwise write.

Whether we decide on arrow functions or classic functions, I think there are linter rules that we can enable to make sure we are consistent in our preference.

src/utils/generateEnumTypeguard.ts
jasonaowen: (Yes, @slifty has rewritten the function for clarity; this conversation is on an outdated diff, which I elected to continue despite the code under discussion being rewritten so as to avoid splitting the conversation. See commit https://github.com/PermanentOrg/permanent-sdk/pull/14/commits/4b1cb04383e715c936f59a655d3d71881cd7e904.)

src/utils/archiveVoToThumbnails.ts
slifty: oh SNAP!!!!!!!!!  great catch

slifty:
Ugh, github, you make threaded conversations impossible...

I'm not deeply opinionated here but wanted to float the approach described in [this StackOverflow thread](https://stackoverflow.com/a/23045200/159522).

Essentially it suggests:

```
1. Use `function` in the global / module scope and for Object.prototype properties.
2. Use class for object constructors.
3. Use => everywhere else.
```

Though I would add

```
4. use `function` for class methods that don't access `this` (I think this is in the linting rules already)
```

The post talks about why in more detail, but the argument for functions in the global scope that compel me are:

1. We actually don't want `this` to be used or accessible for functions outside of a class.
2. They are named and the name shows up in stack traces.

There's also a very minor "`function` makes it super explicit that we're defining a functon" but that's not gonna be the most compelling I suspect.

Does this all make sense?  I'll at the very least go through and make sure I'm consistent (I forgot to convert the enum typeguard to this pattern, for instance).  Maybe we can talk about it quickly at the end of our planning today.

src/utils/createArchiveVo.ts
slifty: My initial reaction was 🤔 😲 

But then I thought about it and it became 🤔 😀 

I think I like this idea; it meshes with the idea that we aren't using models -- there is no need for a factory when we can just "create" an object by cloning the defaults.

It's not a pattern I'm used to, but it does seem elegant enough.

src/utils/archiveVoToThumbnails.ts
slifty: Oh cool -- I hadn't seen this technique before, very nice ty!

slifty: Liner isn't pleased; unfortunately this would result in an array that includes `undefined` elements... I could add a filter, but this is starting to feel a little dirty.  I wonder if there's a way to just return a truly "noop" type of thing.

image

slifty: This article talks about a way of using the spread operator -- https://2ality.com/2017/04/conditional-literal-entries.html

It mentions the downside is that things become a little cryptic, which I can understand.

I do worry that simple arrow functions (as opposed to breaking things into parts) becomes less legible for strings of complex operations but appreciate that it depends.

slifty: https://stackoverflow.com/questions/44908159/how-to-define-an-array-with-conditional-elements

This has some options too; I do worry this starts to get to be one of those "too clever" code approaches.

src/utils/generateEnumTypeguard.ts
jasonaowen: @jvasile confirmed elsewhere that he is not concerned about this, and that I'm making much ado about nothing. Thanks for going down this too-careful licensing road with me, everyone!

src/utils/archiveVoToThumbnails.ts
jasonaowen: After pairing for a few minutes, we came up with a much better option! The `if` statements are protecting against null but defined `thumbURL`s; instead, we can unconditionally include them, and add a `filter` step before the final `map`.

slifty: OK!

We briefly paired on this question over slack and the result is *so much nicer than anything before seen by human kind*.

```
  [
    [200, archiveVo.thumbURL200],
    [500, archiveVo.thumbURL500],
    [1000, archiveVo.thumbURL1000],
    [2000, archiveVo.thumbURL2000],
  ].filter(([_, url]) => url !== null)
```

src/utils/createArchiveVo.ts
jasonaowen: > That said, we are going to be exposing this pattern to end users who might be more comfortable with a utility function

Ah, I hadn't realized that! I had hoped we'd be able to protect the users from the VOs. When would they need to construct one directly?

src/utils/archiveVoToThumbnails.ts
slifty: Ugh (noting this here just to share).

image

Why must I always find typescript type detection edge cases.

It's upset because even though my values are all [number, string] it's mapping them to (number | string)[]...

slifty:
After a whole bunch of discussion and research here's what I would like to suggest:

1. Default to arrow functions (`=>`)
2. Never use arrow functions for methods, [as documented in the MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions).  This means class methods (static and instance) should be defined as functions.
3. If we need to use recursion or a generator function, use functions.

Some guides mentioned that arrow functions should not be used for things like DOM event callbacks because `this` is dynamically populated to reference the event.  However: that same value can be accessed more explicitly and clearly through the callback parameters.  This means that for frontend situations where event handlers are NOT class methods, they should be arrow functions.

We did look at a ton of other library examples, which tended to use the `function` keyword.  It is not obvious exactly why they aren't using arrows, but we strongly suspect it is because *they were started before arrow functions existed / fully appreciated* -- also because many libraries want to minimize reliance on transpiling for compatibility.  We are not burdened with the transpile problem.

`function` has a few challenges like the ones Jason mentioned; it also has a risk of being silently overwritten or deleted.  Arrow functions are stored as `const` which modification would throw a loud error.

Note that the intent is for this to be a set of rules that we can safely use across all of Permanent, not just this repository.

src/utils/generateEnumTypeguard.ts
```diff
@@ -0,0 +1,10 @@
+// This is a modified version of code written by StackOverflow user @jcalz
+// Copyright CC BY-SA 4.0
```
slifty, line 2: I apologize for revisiting but -- is the verdict that I shouldn't add this copyright any longer (I believe this license doesn't work with #15 as it stands)

jasonaowen: My understanding - and I checked with @kfogel, who agrees - that no, we shouldn't add this copyright any longer.

src/utils/archiveVoToThumbnails.ts
```diff
@@ -0,0 +1,39 @@
...
+import { ThumbnailStatus } from '../types';
+import { isThumbnailStatus } from './isThumbnailStatus';
+
+const statusCodeToThumbnailStatus = (statusCode: string): ThumbnailStatus =>
+  isThumbnailStatus(statusCode) ? statusCode : ThumbnailStatus.Unknown;
+
+export const archiveVoToThumbnails = (archiveVo: ArchiveVo): Thumbnail[] => {
+  const thumbnailStatus: ThumbnailStatus = statusCodeToThumbnailStatus(archiveVo.thumbStatus);
+  const thumbnailCreatedAt = new Date(archiveVo.thumbDT);
+  return [
```
jasonaowen, line 15: I swear I'm not just code golfing, but I do honestly think this could be more readable as an expression! The variables are only used once, and so could be inlined, which would lead to shorter lines (particularly for the status). Then all that's left is the `return` statement, which means the arrow function no longer needs a multiline body.

src/utils/archiveVoToThumbnails.ts
```diff
@@ -0,0 +1,39 @@
...
+      'resolution': 500,
+      'url': archiveVo.thumbURL500,
+    },
+    {
+      'resolution': 1000,
+      'url': archiveVo.thumbURL1000,
+    },
+    {
+      'resolution': 2000,
+      'url': archiveVo.thumbURL2000,
```
jasonaowen, line 30: Why quote `'resolution'` and `'url'` here? Should we do the same with `status` and `createdAt`, below? Or not, in both places?

src/utils/__test__/archiveVoToThumbnails.test.ts
```diff
@@ -0,0 +1,57 @@
...
+  it('should handle unknown thumbnail statuses', () => {
+    const archiveVo = {
+      ...defaultArchiveVo,
+      thumbDT: '2022-02-27 10:00:00',
+      thumbStatus: 'this.is.not.a.real.status',
+      thumbURL200: 'example.com/200',
+    };
+    const thumbnails = archiveVoToThumbnails(archiveVo);
+    expect(thumbnails.length).toBe(1);
+    expect(thumbnails[0].status).toEqual(ThumbnailStatus.Unknown);
```
jasonaowen, line 55: Love the arrange-act-assert structure of all these tests!

src/utils/__test__/archiveVoToThumbnails.test.ts
```diff
@@ -0,0 +1,57 @@
...
+      thumbURL200: 'example.com/200',
+      thumbURL2000: 'example.com/2000',
+      thumbURL500: 'example.com/500',
+    };
+    const thumbnails = archiveVoToThumbnails(archiveVo);
+    expect(thumbnails.length).toBe(4);
+    thumbnails.forEach((thumbnail: Thumbnail) => {
+      expect(thumbnail.status).toEqual(thumbnailStatus);
+      expect(thumbnail.createdAt).toEqual(thumbnailCreatedAt);
+    });
```
jasonaowen, line 30: Do we want to test that the URLs are correctly transferred to the thumbnails? This might be a place where we can be a little bit clever, like

```typescript
expect(thumbnail.url).toEqual(`example.com/${thumbnail.resolution}`);
```

since that's only encoding knowledge about the test structure rather than what we expect from the API.

jasonaowen (approved):
Awesome work, @slifty! This looks fantastic, and I really appreciate the care and consensus-building approach you've taken in this pull request. Thank you so much!

(below are optional, non-blocking, and I'm completely fine with merging this as-is)

src/utils/archiveVoToThumbnails.ts
slifty: So I *did* consider that I swear -- the reason I opted for this was because though the variables are only used once from a *lines of code* POV, the line that uses the variables is invoked once per element, which means we would be doing the conversion once per array element.

We're talking 4x right now so it doesn't much matter, but it made me very morose to think about those wasted calls.

src/utils/archiveVoToThumbnails.ts
slifty: Good call -- I don't know why the linter didn't yell about this.

src/utils/__test__/archiveVoToThumbnails.test.ts
slifty: Good suggestion -- will add that in.

src/utils/archiveVoToThumbnails.ts
```diff
@@ -0,0 +1,39 @@
...
+} from '../types';
+
+import { ThumbnailStatus } from '../types';
+import { isThumbnailStatus } from './isThumbnailStatus';
+
+const statusCodeToThumbnailStatus = (statusCode: string): ThumbnailStatus =>
+  isThumbnailStatus(statusCode) ? statusCode : ThumbnailStatus.Unknown;
+
+export const archiveVoToThumbnails = (archiveVo: ArchiveVo): Thumbnail[] => {
+  const thumbnailStatus: ThumbnailStatus = statusCodeToThumbnailStatus(archiveVo.thumbStatus);
```
slifty, line 13: There is no need to specify type here as far as I can tell (whyyyy doesn't the linter catch that?)

src/utils/archiveVoToThumbnails.ts
jasonaowen: Oh, is that true? That sounds super plausible; I might not have as strong a mental model as I ought to for that kind of runtime consideration.

I also wonder if TypeScript or Babel are able to optimize that in practice, since you're right that those would be needless repetitions. I would have faith that javac or gcc could, but I haven't thought as deeply about that kind of thing in the JavaScript world!

slifty: I checked if babel optimizes when transpiling (I'm curious...) -- I set up statusCode to call in the loop.

It becomes:

```
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
exports.archiveVoToThumbnails = void 0;
const types_1 = require("../types");
const isThumbnailStatus_1 = require("./isThumbnailStatus");
const statusCodeToThumbnailStatus = (statusCode) => (0, isThumbnailStatus_1.isThumbnailStatus)(statusCode) ? statusCode : types_1.ThumbnailStatus.Unknown;
const archiveVoToThumbnails = (archiveVo) => {
    // const status = ;
    const createdAt = new Date(archiveVo.thumbDT);
    return [
        {
            resolution: 200,
            url: archiveVo.thumbURL200,
        },
        {
            resolution: 500,
            url: archiveVo.thumbURL500,
        },
        {
            resolution: 1000,
            url: archiveVo.thumbURL1000,
        },
        {
            resolution: 2000,
            url: archiveVo.thumbURL2000,
        },
    ].filter(thumbnail => thumbnail.url !== '')
        .map(thumbnail => ({
        ...thumbnail,
        status: statusCodeToThumbnailStatus(archiveVo.thumbStatus),
        createdAt,
    }));
};
exports.archiveVoToThumbnails = archiveVoToThumbnails;
```

Which yeah, means it's just gonna call it once per loop :(

PR #15, opened 2022-03-08
Change license to MIT

kfogel:
As discussed with Robert Friedman and Cecilia Krum.


slifty:
Looks good -- I would personally love to know the reason for MIT instead of LGPL just for my own edification, but I'm happy to get filled in outside of the PR.

cecilia-donnelly (approved):
Thanks @kfogel !

cecilia-donnelly:
@slifty I think @kfogel suggested MIT because it's common in javascript land but I will let him answer!

kfogel:
Robert had a good way of saying it (I can only paraphrase -- I wish I'd recorded him, but I'm not, uh, in one of those one-party-notification states, you know):

The Permanent.org platform is AGPL because it's for the general good of humankind, and therefore if anyone's running an instance of this service we want to encourage them to contribute to that same general good.  The purpose of the SDK is subtly different: the SDK's job is to make it easy for everyone to interact with the platform that provides this general good.  The type of licensing that best serves this purpose is non-copyleft, because that enables Permanent-reaching technology to be integrated into as many applications as possible.

slifty:
Good to know!  I asked because I was under the impression that LGPL would NOT be copyleft in the `import` case / for people using it -- and the copyleft would only apply to modifications to the SDK itself.

This does mean someone could fork our work here on this SDK and incorporate what we create in closed projects (up to and including a closed source permanent SDK), right?

kfogel:
@slifty That's right, and we're okay with that.

Issue #20, opened 2022-03-21
Publishing to npm


slifty:
We want to publish to npm.

Doing this is straightforward but I'd like to pair it with establishment of the following:

1. [A changelog](https://keepachangelog.com/en/1.0.0/) + adherence to [semver](https://semver.org/spec/v2.0.0.html)
2. A cycle of thinking around CD for deployment

It also means doing a quick pass to ensure we're publishing exactly what we want to publish (npm has the concept of an .npmignore file which we can use if there's anything we want to ignore from the published package).

slifty:
Regarding CI, this guide shows how to set up a github action to publish to npm when a new release is created:

https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages

slifty:
Also note that github releases are done via the UX / not command line.

There are some tools for creating github releases via command line; we could write our own custom script as well.

https://stackoverflow.com/questions/21214562/how-to-release-versions-on-github-through-the-command-line

That said, starting releases as a manual process might be fine -- we will have to do the following:

1. update the changelog (add a verison header / links, move items to the new header)
2. update the package.json version
3. (open a PR for those changes?)
4. "Create a release" via the github UX once that PR is merged.

It would be nice if github made it possible to combine step 3 and 4.
It would be nice if we could find a way to automatically do step 1 and 2.

slifty:
Just a note: we currently transpile to `lib` -- this is a *little* unclear since `lib` can also mean "third party libraries used by the code".  There's no real standard [though this SO answer goes into it a bit](https://stackoverflow.com/questions/22842691/what-is-the-meaning-of-the-dist-directory-in-open-source-projects)

I'm going to rename our build location to `build`

`dist` is also an option and I only landed on `build` because... idk I guess it makes me feel like a real programmer using a real programming language.  (the point I'm trying to make is that if anybody would prefer `dist` that's fine by me)

(I reserve the right to edit this post and completely change the decision)

slifty:
[Here's a useful article](https://blog.theodo.com/2021/04/library-tree-shaking/) about making sure your published library supports tree shaking.  I am going to avoid spending too much time on this right now since tree-shaking a library that offers zero functional code isn't that useful, but we *are* going to want to support it so if it's easy to do it up front then great.

slifty:
OK... our SDK does soemthing; I'm going to see if this code from March was actually correct!

slifty:
Update: it was!  Closing this issue.

PR #21, opened 2022-03-22
Support NPM publication

slifty:
This PR adds logic for (dry-run) publishing to NPM whenever we release a new github version.

I'm starting this with --dry-run enabled because I don't want to *actually* publish anything until we're sure our github hooks are properly configured.  Once that's done I'll open a new PR to remove the `--dry-run` flag.

Part of the changes involve changing what and where we build for distribution.

Related to #20

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,18 @@
+name: publish (npm)
+on:
+  release:
+    types: [created]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+      - run: npm run build
+      - run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
--- a/.gitignore
+++ b/.gitignore
@@ -102,6 +102,3 @@ dist
 
 # TernJS port file
 .tern-port
-
-# Built directory
-lib/
--- a/package.json
+++ b/package.json
@@ -2,10 +2,13 @@
   "name": "permanent-sdk",
   "version": "0.0.0",
   "description": "Functional Node.js SDK for Permanent.org",
-  "main": "src/index.ts",
+  "module": "dist/esm/index.js",
+  "main": "dist/cjs/index.js",
   "scripts": {
     "lint": "eslint ./src --ext .ts",
-    "build": "tsc",
+    "build": "npm run build:node && npm run build:browser",
+    "build:node": "tsc -p tsconfig.node.json",
+    "build:browser": "tsc -p tsconfig.browser.json",
     "test": "jest"
   },
   "repository": {
@@ -31,6 +34,7 @@
   },
   "homepage": "https://github.com/PermanentOrg/permanent-sdk#readme",
   "devDependencies": {
+    "@tsconfig/esm": "^1.0.0",
     "@tsconfig/node14": "^1.0.1",
     "@types/jest": "^27.4.0",
     "@typescript-eslint/eslint-plugin": "^5.10.0",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "permanent-sdk",
+  "name": "@permanentorg/sdk",
   "version": "0.0.0",
   "description": "Functional Node.js SDK for Permanent.org",
   "module": "dist/esm/index.js",
@@ -15,6 +15,7 @@
     "type": "git",
     "url": "git+https://github.com/PermanentOrg/permanent-sdk.git"
   },
+  "funding": "https://permanent.org/donate/",
   "author": {
     "name": "Permanent.org",
     "email": "engineers@permanent.org",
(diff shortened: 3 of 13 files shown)


.gitignore
```diff
@@ -104,4 +104,4 @@ dist
 .tern-port
 
 # Built directory
-lib/
```
jasonaowen, line 107: `dist` is already ignored on line 83! I don't think you need to make any changes to this file.

package.json
```diff
@@ -29,9 +38,14 @@
   "engines": {
     "node": ">=14.0"
```
jasonaowen, line 39: Shall we update this to 16, too?

jasonaowen (commented):
Do we also need an `.npmignore` file? Last time I did something like this, I seem to recall that the best practice was to ignore input TypeScript files and only package the build artifacts, and by default `npm pack` and `npm publish` use `.gitignore` in the absence of `.npmignore`, which results in the opposite.

slifty:
We don't need `.npmignore` -- npm ignore is only needed if you don't have [`files` configured](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#files) in `package.json` (the ultimate fallback is `.gitignore` which we don't want of course since that ignores the `dist` directory!)

package.json
slifty: I am realizing maybe I was overeager in moving the tsconfig to 16.

14 is still around for a while and I think we wanted to support node versions in maintenance mode.

I think I should reverse that change and use node14 in the tsconfig.

package.json
```diff
@@ -29,9 +38,14 @@
...
   },
+  "files": [
+    "dist",
+    "src"
```
slifty, line 43: breaking this out of the non-threaded conversation UX (ugh github)

@jasonaowen I would normally not include `src` but wasn't sure if we would want to in order to support people who like inspecting code bases.

You cool with me taking it out?

.gitignore
jasonaowen: I was going to say that this file originally came from https://github.com/github/gitignore/blob/ce5da10a3a43c4dd8bd9572eda17c0a37ee0eac1/Node.gitignore and that there's merit in keeping it similar, but it appears we added `lib` and upstream doesn't have it - so yes agree!

package.json
jasonaowen: I am cool with you taking it out, yes! Folks can inspect the somewhat-readable compiled JavaScript, the very-readable type definitions, or go directly upstream to this repo to see the originals.

package.json
```diff
@@ -1,17 +1,21 @@
...
-  "main": "src/index.ts",
+  "module": "dist/esm/index.js",
+  "main": "dist/cjs/index.js",
   "scripts": {
     "lint": "eslint ./src --ext .ts",
-    "build": "tsc",
+    "build": "npm run build:node && npm run build:browser",
+    "build:node": "tsc -p tsconfig.node.json",
+    "build:browser": "tsc -p tsconfig.browser.json",
     "test": "jest"
```
jasonaowen, line 12: (Picking a semi-arbitrary line to start a threaded conversation on...)

I see this message in the CI logs:

> ts-jest[config] (WARN) message TS151001: If you have issues related to imports, you should consider setting `esModuleInterop` to `true` in your TypeScript configuration file (usually `tsconfig.json`). See https://blogs.msdn.microsoft.com/typescript/2018/01/31/announcing-typescript-2-7/#easier-ecmascript-module-interoperability for more information.

I don't have a quick answer on what to do here, and I haven't dug into that blog post. Should we have a TSConfig for Jest? Or point Jest at one of these three configs?

slifty: Argh -- I bet jest auto-loaded `tsconfig.json` when it was named that, and by splitting it into two pieces I caused that issue.  I'll look into this.  Maybe the answer is just renaming `tsconfig.lint.json` to `tsconfig.json`.

slifty: Update: yeah that fixes it.

OK so: I don't like that jest was doing that magic in a way that requires a weird `tsconfig.json` name so I'm going to:

1. Rename the lint version to `tsconfig.dev.json` <-- which I think is an appropriately specific-to-lint-and-test-but-generic-to-cover-both
2. Make the tsconfig setting explicit in the `jest.config.js`

jasonaowen: That sounds great! Awesome, thank you @slifty!

.github/workflows/release.yml
```diff
@@ -0,0 +1,18 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: '16.x'
+          registry-url: 'https://registry.npmjs.org'
+      - run: npm ci
+      - run: npm run build
+      - run: npm publish --dry-run
```
jasonaowen, line 16: Would it make sense to also do this dry run as part of CI on pull requests?

jasonaowen (approved):
These changes all make sense to me! Nicely done, @slifty.

I haven't tested this PR; I believe @meisekimiu has a test plan, and even aside from that I look forward to reading her take on this.

.github/workflows/release.yml
slifty: Oh brilliant, yes.

We can also do it on merge to main -- that way we know exactly what the package is going to contain before issuing a release without having to look around for whatever the most recent PR was.

slifty:
Side note: I wonder if codecov didn't trigger because I made no changes to code, or if it didn't run because it's broken already.

Issue #17, opened 2022-03-09
Create folder types + utilities


slifty:
The `Folder` and `FolderVo` types are empty right now and need to be populated.

We also need to create utilities for converting from `folderVoToFolder`

slifty:
And here I was thinking `ArchiveVo` was fun to figure out!!

[FolderVo](https://github.com/PermanentOrg/back-end/blob/main/api/core/folder/vo/folder.vo.php) has a lot going on, so I want to document some of my exploration.

There are some public attributes and some private attributes on the base class; it isn't clear what the actual footprint of what is returned by the API will look like.

The API provides a `api/folder/get` and a `api/folder/getWithChildren`

My next step is going to be to just use the API directly and see what comes back.

PR #24, opened 2022-04-15
Define API type `FolderVo`

Fenn-CS:
This PR populates the`FoldeVo` type and creates the utilities to covert from the API `FolderVo` to sdk `FolderVo`.

- [x] Create other types `FolderVo` depends on
- [x] Populate `FolderVo` type

Fixes : #17

### DIFF
--- a/src/types/api/FolderVo.ts
+++ b/src/types/api/FolderVo.ts
@@ -1,3 +1,16 @@
-// This is a placeholder interface which will ultimately be populated
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface FolderVo {}
+export interface FolderVo {
+  // These fields map to those defined in the base library ArchiveVO
+  // https://github.com/PermanentOrg/back-end/blob/main/api/core/folder/vo/folder.vo.php
+  folderId: number;
+  archiveId: number;
+  displayName: string;
+  displayDT: string;
+
+}
+
+export const defaultFolderVo: FolderVo = {
+  folderId: 0,
+  archiveId: 0,
+  displayName: '',
+  displayDT: '',
+};


slifty:
@Fenn-CS this isn't a full review yet but just to say this is looking like an excellent start -- if you decide that you want to just do a PR for the FolderVO types (as opposed to a PR that also includes the folder types) I think that's fine.  Either approach works, of course.

I'll aim to give this a closer look shortly.

src/types/api/FolderVo.ts
```diff
@@ -1,3 +1,93 @@
...
+import type { ShareVo } from './ShareVo';
+import type { AccessVo } from './AcessVo';
+
+export interface FolderVo {
+
+  // These fields map to those defined in the base library ArchiveVO
+  // https://github.com/PermanentOrg/back-end/blob/main/api/core/folder/vo/folder.vo.php
+  
+  folderId: number;
+  archiveNbr: number;
```
jasonaowen, line 18: Unfortunately, `archiveNbr` is not a number, despite its name; see the [discussion in PR #10](https://github.com/PermanentOrg/permanent-sdk/pull/10#issuecomment-1033181992) for more information.

src/types/api/FolderVo.ts
```diff
@@ -1,3 +1,93 @@
...
+  parentArchiveNbr: number;
+  parentDisplayName: string;
+  pathAsArchiveNbr: string;
+
+  //Children
+  ChildFolderVOs: FolderVo[];
+  RecordVOs: RecordVo[];
+  LocnVO: LocationVo;
+  TimezoneVO: TimezoneVo;
+  DirectiveVOs: DirectiveVo;
```
jasonaowen, line 75: Oh geez... uhh can we ignore these? They need to go away!

src/types/api/FolderVo.ts
```diff
@@ -1,3 +1,93 @@
...
+  archiveNbr: number;
+  archiveId: number;
+  displayName: string;
+  downloadName: string;
+  displayDT: string;
+  displayEndDT: string;
+  derivedDT: string;
+  derivedEndDT: string;
+  
+  note: string;
```
jasonaowen, line 27: I have no idea what a `note` is, so maybe this is a great one to comment on: I think many of these need to allow for `null`s, for two reasons: the API might not include it in responses, and the client might not be able to provide it in requests. Compare [web-app's `FolderVO`](https://github.com/PermanentOrg/web-app/blob/05adce6d9c42512ab27c361f8f9d7dccce2c1128/src/app/models/folder-vo.ts), which leaves many of these as implicit `any`s - let's not do that, but I think we do need some additional flexibility.

It might be necessary or desirable to split these definitions into `RequestFolderVO` and `ResponseFolderVO`; that'd allow for encoding into type definitions information like "creating a new folder does not allow for a `folderId`, but folders returned from the API will always have a `folderId`".

jasonaowen (commented):
This is not a full review, but hopefully some useful conversation starters!

src/types/api/FolderVo.ts
slifty: @jasonaowen Just wanted to share that my gut reaction was that I don't love the idea of defining a type that allows for null since it puts a bunch more burden on the user of the type to constantly check for presence before they can use an attribute (isn't vanilla JavaScript just TypeScript where all object attribute are nullable and need to be checked before use, at the end of the day... I kid I kid.)

In that no-nullable-attribute world we solve the creation problem through use of `Partial`, `Pick`, and `Omit` when defining functional parameters.

I realize that may be backwards though, and maybe the best pattern is to have nullable attributes and avoid Partial / Pick / etc.  I'd like to  into how other TS sdks handle this kind of thing.

Side note, if we do allow "null" the TypeScripte guidelines suggest [using undefined and never null](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined)

Maybe as a first pass we should make sure to eliminate any fields that truly don't belong, then we can see how many nullable fields are left.

(In this case, for example, I don't fully understand why null is preferable to an empty string.)

src/types/api/FolderVo.ts
slifty: Yeah -- and for the record I believe there is a bug in the `ArchiveVo` definition of this repo that I will eventually need to fix.

Fenn-CS:
Thanks @jasonaowen and @slifty! Acting on those later.

src/types/api/FolderVo.ts
jasonaowen: > I don't love the idea of defining a type that allows for null (whenever we can help it) since it puts more burden on the user of the type to constantly check for presence before they can use an attribute

@slifty yes, I completely agree! And that was our goal in building out the user-facing types in #10 and so on. My understanding is that these are the API-facing types, and there we have to deal with the actual requests and responses the API requires and returns.

Several of these fields will sometimes or always have a value of `null` in the API response JSON, and several of these fields can, should, or must be omitted in the JSON sent in API requests. Our type system should properly capture that, or else we'll have runtime errors: if our type definition says a `note` is always a string, and the API gives us a `null` instead, then when client code tries to use the `note` then it'll get a JavaScript `TypeError`; or, if the client tries to defend against that possibility, TypeScript will complain about needless type checks because we've told it the value can never be `null`.

> ...we solve the creation problem through use of the utility types...

Yes, I agree with that, too! Those sound like great mechanisms for implementing these types. In that case perhaps the pattern is define the fullest possible type for each VO, and then take subsets of that type for its request and response forms.

> Maybe as a first pass we should make sure to eliminate any fields that truly don't belong, then we can see how many nullable fields are left.

I'm not sure if you're thinking on the server side or on the SDK side, but I support both! I'm honestly kind of wondering if we shouldn't take a more minimalist approach in the SDK of only defining the properties we actually care about, and adding on to them as we gradually care about more things? That would give us some flexibility on the API side to remove unused fields, knowing that clients of this SDK will not be affected.

jasonaowen:
@slifty and @Fenn-CS will pair on this PR this week.

Fenn-CS:
Some fields are no longer used at the level of the back-end and so are not including in the `FolderVo` type definition of the sdk, here is a list of what we removed for reference:

- `note`
- `LocnVO` 
- `TimezoneVO`
- `DirectiveVOs`

src/types/api/FolderVo.ts
slifty: Where we landed on this was that we are going to specify default values (like we do with `archiveVo`) and worry about the null case at the point api-response-to-typed-object-population.

We may find on implementation that things truly must be nullable.

We will also, at that point of implementation, be able to remove truly unused attributes with more intention.  It will be far easier to understand what is and is not populated when we're actually working with API response data.

src/types/api/FolderVo.ts
```diff
@@ -1,3 +1,123 @@
...
+import type { FolderSizeVo } from './FolderSizeVo';
+import type { ShareArchiveVo } from './ShareArchiveVo';
+import type { ShareVo } from './ShareVo';
+import type { AccessVo } from './AcessVo';
+
+export interface FolderVo {
+
+  // These fields map to those defined in the base library ArchiveVO
+  // https://github.com/PermanentOrg/back-end/blob/main/api/core/folder/vo/folder.vo.php
+
```
slifty, line 15: Take out this whitespace

jasonaowen:
So this is a thought I wish I had come up with three or four weeks ago rather than after there's already one approval on this PR, but...

In this and #26, we appear to be taking a sort of kitchen-sink approach of defining everything we think might be relevant, and planning to modify or delete as we learn more. My concern is that we don't have any code that is using any of these attributes, on either side (parsing responses from the API, or clients of this SDK trying to get things done), so we don't really know what should be included and should be excluded. This will contribute to the unused code problem we're already struggling with, both by adding new unused code and by making it harder to delete unused API code that affects API responses.

The approach I would love for us to adopt, at least going forward, is driving these type definitions from SDK-client code: the sftp-service (or the web-app, or a partner, or...) needs to accomplish a specific, concrete goal. We use that goal to figure out which client-facing models and which attributes on those models we need, then figure out how to translate between the client models and the API models, and only then specify types for the relevant JSON attributes in the API's requests and responses.

For example: the sftp-service needs to know about the directory structure of an archive. That means we need the `Folder` model in #26. We figure out what specifically we need there - `archiveId`, `downloadName`, `createdAt`, etc - AND write the code to populate that information using the `FolderVO` responses, which we then need to define. All of that is closely related: we don't define any new attributes until we know how to use them, and we know we know how to use them because we introduce working code alongside those attributes that does use them.

slifty:
@jasonaowen that approach does make sense ty for articulating it so effectively!

Under the above proposal all we would need are:

1) Fields that are required for creation / deletion of folders and records
2) Fields that allow us to crawl an archive's directory structure
3) Fields that allow us to access the content of a file

Heck, we don't even need thumbnails, statuses, or really anything beyond a small handful of fields (should we delete those from Archive, by the way?)

This would mean we don't need to be asking or thinking about what `SpecialFolderType` is (or `FolderType` for that matter).

Just because it's important to talk through, one challenge is that it puts some burden on the mythical SDK user defining what it is they need in terms of what the API can provide without knowing what the API can provide (both because the API is closed source and also because even with source access it is quite opaque even if you have the files).

This would also mean that we're not as close to being ready for the web app to leverage this SDK / Natalie would have to open a parallel set of PRs to us before she could swap out this SDK for node-sdk (since we'd only be defining a fraction of the attributes webapp needs).  That said: without her integration we have no reason to believe that anything we're writing today would work for her in the first place!

TL;DR -- what you're proposing makes sense and will be faster to implement.  It does make the division of labor a bit less necessary (e.g. after we finish these Folder definitions I should probably take on the File definitions so I'm defining them as part of the sftp-work)

Thoughts all of the above @Fenn-CS?

Fenn-CS:
Yes @jasonaowen that point makes sense but the burden of figuring out how something should work from the code perspective is  much so, it would make sense to work with the API Guide instead : https://permanent.atlassian.net/wiki/spaces/EN/pages/793051137/API+Guide

Since there's nothing like OPEN API annotations, but the challenge with that is, even those definitions are not complete for all the VOs.

A good middle ground would be to practive Jason's approach on the sdk types, it makes complete sense as there's no need defining something that we don't know how to use inside the SDK yet, but for the backend or SDK VO translations I think that the most effective to eliminate unused properties, is to FIRST, remove them from backend and if they are already defined in the SDK, we file a corresponding PR to remove them as well here.

jasonaowen:
That's a good summary, thank you, @slifty!

> Heck, we don't even need thumbnails, statuses, or really anything beyond a small handful of fields (should we delete those from Archive, by the way?)

Deleting them starts getting into the kind of thrash I would like to avoid; but, if attributes are defined but not used or populated by the SDK, then yeah I'd +1 a PR removing them.

> one challenge is that it puts some burden on the mythical SDK user defining what it is they need in terms of what the API can provide without knowing what the API can provide

That's true, but thankfully the current users of this SDK aren't mystical - they're us! We have several specific scenarios that I think we could round up and flesh out without too much effort. Broadly: what sftp-service needs, what the etherpad plugin needs, what web-app needs, and what we know partners are using the node-sdk for. Getting access to the API, as would be needed to effectively use the code in this repo, is a high-touch interaction with the Permanent staff, and we can gather high-level desires and translate them into requirements like "the SDK needs to allow setting the location on a record" or whatever.

> This would also mean that we're not as close to being ready for the web app to leverage this SDK / Natalie would have to open a parallel set of PRs to us before she could swap out this SDK for node-sdk (since we'd only be defining a fraction of the attributes webapp needs). That said and related to your point: without her integration we have no reason to believe that anything we're writing today would work for her in the first place!

:100: exactly - I'd love for us to be able to start migrating API interaction out of web-app and into this repo, and I think/hope the only blockers to that are 1. publishing to NPM 2. adding this SDK as a dependency to web-app and 3. moving each API interaction one-by-one.

Fenn-CS:
@jasonaowen, given your comment I am not sure what action to take next on this PR because I don't know what fields are used in the SDK yet, so my only source of truth is the back-end.

This does not need to be merged urgently, so I don't mind taking definitive action (request changes) on this if you lay it out. If we should pass on just this one, then please leave an approval to merge.

jasonaowen:
> the burden of figuring out how something should work from the code perspective is much so, it would make sense to work with the API Guide instead : https://permanent.atlassian.net/wiki/spaces/EN/pages/793051137/API+Guide

I'm not sure I follow what you're saying here, @Fenn-CS, can you write more? The API guide is a useful resource, but like all documentation, suffers the risk of being left behind as code changes. Either way, what I'm trying to say is that we should figure out how stuff works *as we need it* rather than *all at once up front*, which is what so many of the discussions on this and #26 are about.

> A good middle ground would be to practice Jason's approach on the sdk types, it makes complete sense as there's no need defining something that we don't know how to use inside the SDK yet, but for the backend or SDK VO translations I think that the most effective way to eliminate unused properties, is to FIRST, remove them from back-end and if they are already defined in the SDK, we file a corresponding PR to remove them as well here.

I'm 1000% onboard with removing unused properties on the back-end; we should absolutely be doing that and I'd love any help you can offer there! Part of my concern about overly-detailed models in this repository is about making that work more difficult; we already need to update the sample responses used in web-app tests when we delete a property or a constant from the back-end, and this would be one more place to do so.

> I am not sure what action to take next on this PR because I don't know what fields are used in the SDK yet, so my only source of truth is the back-end.
> 
> This does not need to be merged urgently, so I don't mind taking definitive action (request changes) on this if you lay it out. If we should pass on just this one, then please leave an approval to merge

Thanks, @Fenn-CS! I'll suggest a course of action, but you and @slifty should chat and come to an agreement about how to proceed.

I think this PR and PR #26 are each 1/3 or 1/4 of a complete pull request. I suggest closing one and reworking the other to include both the SDK-client-facing and API-facing models (eg both `Folder` and `FolderVO`), as well as writing the code needed to convert between them in at least one direction and the code to call an API endpoint that uses these models. That will give us runnable, testable, usable code, and we can talk about the models, conversions, and so on with full context and be able to answer questions such as "do we need this?" with "yes, we use it on line X" or "no, not yet, because we're not using it anywhere".

Does that make sense? How does that sound?

Fenn-CS:
> I'm not sure I follow what you're saying here, @Fenn-CS, can you write more?

I meant that for someone like me who has not used the sdk or written some functionality within it before now, even "figuring out what's needed for the moment is challenging"...

That was another way to say that, I don't know what properties to leave here.

But your concluding suggestion gives me the impression that, we can't know by just observing these properties.

> I think this PR and PR https://github.com/PermanentOrg/permanent-sdk/pull/26 are each 1/3 or 1/4 of a complete pull request. I suggest closing one and reworking the other to include both the SDK-client-facing and API-facing models (eg both Folder and FolderVO), as well as writing the code needed to convert between them in at least one direction and the code to call an API endpoint that uses these models. That will give us runnable, testable, usable code, and we can talk about the models, conversions, and so on with full context and be able to answer questions such as "do we need this?" with "yes, we use it on line X" or "no, not yet, because we're not using it anywhere".


Yest that makes sense, actually that was the goal, we pieced the PRs to simplify things. I had the following steps:

- [ ] Create types API Facing `FolderVo` depends on
- [ ] Populate API Facing `FolderVo` type
- [ ] Populate SDK-client-facing `Folder` and create new custom types that should map from `FolderVo`
- [ ] Create utility to convert from API-facing `FolderVo` to SDK-client-facing `Folder` and vice-verser.
- [ ] Call API endpoint to use models (this is the only step that was missing in the plan)

So perhaps, we can leave both PRs open for the same reason and possibly branch out more PRs from 26 and then from the final branched-PR we can test on that branch because it carries all this early work and know what to change in this early PRs. @jasonaowen 


@slifty what do you say?

jasonaowen:
> I don't know what properties to leave here. But your concluding suggestion gives me the impression that, we can't know by just observing these properties.

Ah, yes exactly - the only way to know is by doing!

> So perhaps, we can leave both PRs open for the same reason and possibly branch out more PRs from 26 and then from the final branched-PR we can test on that branch because it carries all this early work and know what to change in this early PRs.

The challenge that approach carries is the one we're discussing here: we cannot judge whether the changes in this PR, or in #26, or in the proposed future PRs, are good and correct because each is incomplete. Viewing each of those partial pieces of work leads to a bunch of comments like "is this the right thing to do here? I can't tell because I don't have enough context".

I get the sense that you are concerned about opening overly-large pull requests, and that is an excellent thing to worry about. I think that concern is addressed, at least in part, by being more conservative with defining models. #24 and #26 add about 400 lines of code between them; but, if we decided that for a first pass, we only really care about say half-a-dozen attributes of `FolderVO` and their translations to `Folder`, that would be a much smaller change - even with the additional code to do those translations.

jasonaowen:
Talking with @slifty, I am reminded of something I wanted to say: the work y'all have done on these PRs *is* good and useful, even if we were to close both without merging. First, this is a fantastic exercise for learning how the API works, and has already prompted several great discussions; and second, while the approach I'm advocating is for is more minimal, it is incredibly valuable to step back and look at the big picture to make sure we're going in the right direction and not designing ourselves into a corner. Thank you both, @Fenn-CS and @slifty!

Fenn-CS:
I and Dan a follow up, meeting decided a good path is to start by defining what we know for sure we would use including what we know is required at the back-end.

So this PR has been super simplified to reflect this path. @jasonaowen this means, that we would not have to hit any back-end API to finish this.

We would add new properties when we hit the back-end API and it ask for specific properties.

jasonaowen:
@Fenn-CS thanks for reducing the scope of this! At a glance, that looks like a reasonable set of attributes to start with.

I would still like to see both the API-facing models in this PR, the user-facing models in #26, and the code that translates between them all in one commit. As I said earlier, it's hard to review the changes in this PR as it's currently structured because they are so incomplete. I got the impression from @slifty's comments that he and I were on the same page about that, but I might have misunderstood.

So, concretely, I propose:
- close #26 without merging
- add its changes to this PR
- add code that converts between a `FolderVO` and a `Folder`, in both directions

Does that make sense?

src/types/api/FolderVo.ts
```diff
@@ -1,3 +1,16 @@
-// This is a placeholder interface which will ultimately be populated
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface FolderVo {}
+export interface FolderVo {
+  // These fields map to those defined in the base library ArchiveVO
+  // https://github.com/PermanentOrg/back-end/blob/main/api/core/folder/vo/folder.vo.php
+  folderId: number;
+  archiveId: number;
+  displayName: string;
+  displayDT: string;
```
slifty, line 7: We should also include downloadName

Fenn-CS:
We agreed in a meeting that having this issue was an extra burden for management since it's not too big and not too seperate from #26 where the work in merged would continue to progress from there.

PR #26, opened 2022-05-06
Add SDK definition for Folder

Fenn-CS:
Following the addition of the API Type `FolderVo` in #24 this PR defines the SDK Type in `sdk/types/Folder.ts`

## TODO
- [x] Populate `Folder` and create new custom types that should map from `FolderVo`

Ticker: #17


src/types/sdk/Folder.ts
```diff
@@ -1,3 +1,60 @@
...
+  // thumbStatus: string;
+  // thumbURL200: string;
+  // thumbURL500: string;
+  // thumbURL1000: string;
+  // thumbURL2000: string;
+  // thumbDT: string;
+  status: GenericStatus;
+  publicDT: string;
+  parentFolderId: number;
+  // folder_linkType: string;
```
Fenn-CS, line 32: `FolderLink` is bridge table concept I believe, and so does not need exist here! Please @jasonaowen confirm for removal.

src/types/sdk/Folder.ts
```diff
@@ -1,3 +1,60 @@
...
+  // pathAsFolder_linkId: number;
+  // pathAsText: string;
+  // folder_linkId: number;
+  // parentFolder_linkId: number;
+  // ParentFolderVOs: FolderVo[];
+  // parentArchiveNbr: number;
+  // parentDisplayName: string;
+  // pathAsArchiveNbr: string;
+  // ChildFolderVOs: FolderVo[];
+  // RecordVOs: RecordVo[];
```
Fenn-CS, line 46: - `pathAsFolder` - what is that for?
- `FolderLink` - anything FolderLink should not exists here ideally
- `ParentArchive` - Not required, archive Id exists, if we get the archive we can check the parent of the archive
- `*Archive` - No other `Archive` property except slug and ID should appear here?
- `ChildFolders` and `RecordVOs` - replaced with the `children` prop

src/types/sdk/Folder.ts
```diff
@@ -1,3 +1,60 @@
...
+  // SharedArchiveVOs: ShareArchiveVo;
+  // FolderSizeVO: FolderSizeVo; // Replaced with size simply!
+  // ChildItemVOs: FolderVo[]; // Replaced with children
+  // ShareVOs: ShareVo[];
+  // AccessVO: AccessVo;
+  // AccessVOs: AccessVo[];
+  // archiveArchiveNbr: number;
+  // returnDataSize: string;
+  // posStart: number;
+  // posLimit: number;
```
Fenn-CS, line 57: Anything here needed?

src/types/sdk/Folder.ts
```diff
@@ -1,3 +1,60 @@
...
+import type { AccessRole } from './AccessRole';
+import type { FolderType } from './FolderType';
+import type { GenericStatus } from './GenericStatus';
+import type { FolderSortOrder } from './FolderSortOrder';
+import type { FolderSpecialType } from './FolderSpecialType';
+
 export interface Folder {
+  id: number;
   size: number;
+  archiveSlug: string;
```
slifty, line 13: Given that archiveId is an attribute already I'm not sure if archiveSlug should be part of this type, since archiveId can be used to load the `archive` and then `archive.slug` would be the same as this attribute.

src/types/sdk/Folder.ts
```diff
@@ -1,3 +1,60 @@
...
+  archiveId: number;
+  displayName: string;
+  downloadName: string;
+  createdAt: Date;
+  derivedAt: Date;
+  description: string;
+  special: FolderSpecialType;
```
slifty, line 20: I wonder if we can come up with a better name for this than `special` -- maybe `role` -- and also why is it different from FolderType at all.

This warrants some research / discussion IMHO.

src/types/sdk/Folder.ts
```diff
@@ -1,3 +1,60 @@
...
+  sort: FolderSortOrder;
+  type: FolderType;
+  // thumbStatus: string;
+  // thumbURL200: string;
+  // thumbURL500: string;
+  // thumbURL1000: string;
+  // thumbURL2000: string;
+  // thumbDT: string;
+  status: GenericStatus;
+  publicDT: string;
```
slifty, line 30: I think with archive we removed this attribute.

We do have a `readonly containsPublicContent: boolean;` in archive.

src/types/sdk/Folder.ts
```diff
@@ -1,3 +1,60 @@
...
+  parentFolderId: number;
+  // folder_linkType: string;
+  // FolderLinkVOs: FolderLinkVo[];
+  accessRole: AccessRole;
+  // position: number;
+  shareAT: Date;
```
slifty, line 36: This might be the kind of feature we can just remove from the sdk until there's a use case / need for it.

If we keep it it should be `sharedAt`

src/types/sdk/Folder.ts
```diff
@@ -1,3 +1,60 @@
...
+  // ChildItemVOs: FolderVo[]; // Replaced with children
+  // ShareVOs: ShareVo[];
+  // AccessVO: AccessVo;
+  // AccessVOs: AccessVo[];
+  // archiveArchiveNbr: number;
+  // returnDataSize: string;
+  // posStart: number;
+  // posLimit: number;
+  readonly thumbnails: Thumbnail[];
+  readonly children: (Folder | Item)[];
```
slifty, line 59: One incoming change coming to `Archive` is rolling back the design decision to lump both types of children together.  That just means users of the SDK have to add a bunch of clauses to check the type of each child.

Let's instead have `folders` and `items` as two separate attributes.

src/types/sdk/Folder.ts
jasonaowen: Wait, I thought `archiveSlug` was the more-descriptive name for `archiveNbr`, which we do want the SDK to support for the sake of the web-app?

src/types/sdk/FolderSpecialType.ts
```diff
@@ -0,0 +1,5 @@
+// These types are defined in the permanent backend:
+// https://github.com/PermanentOrg/back-end/blob/main/library/base/constants/base.constants.php
+export enum FolderSpecialType {
+  FacebookRootFolderSpecial = 'facebook.root.folder',
```
slifty, line 4: There are a bunch of special types I believe; from @jasonaowen's query, these are all currently possible values (though many are depricated / not real any more):

```
(null)	80406
(emptystring)	274
dropbox.retrieve.folder	22
dropbox.retrieve.folder2	1
dropbox.root.folder	30
dropbox.sync.folder	20
facebook.root.folder	288
facebook.root.folder2	1
familysearch.root.folder	2434
folder.special.crop	1
folder.special.dropbox	4
folder.special.facebook	6
folder.special.gdrive	1
gdrive.root.folder	17
```

I would suggest we ONLY include types that are intended for use (which might just be `familysearch.root.folder` at this point).

I still don't understand the difference between this and `FolderType` though.

src/types/sdk/Folder.ts
slifty: @jasonaowen yes!  slug is the sdk-replacement-name for `archiveNbr`; we want `slug` to be the archive attribute.

My question here is why do we want to store archive attributes on the folder at all (e.g. we don't store `archiveName` or `archiveDescription` here -- rather, the user would load the archive via a `getArchive` call using `folder.archiveId` as an input parameter).  Does that make sense?

If there's a folder-specific reason for storing it here I'd rather call it something folder-specific (e.g. `rootPath` or something like that)

jasonaowen: Folders ALSO have these things, and they're also called `archiveNbr` in the VO / database. I believe that's what allows the web-app to figure out that the path `/app/myfiles/06r0-000b/271660` maps to "folder 271660 in archive 06ro-000b".

(I'm not 100% certain on the mechanics of that; tagging @meisekimiu in case she has corrections or additional insight)

slifty: <s>omg are you serious.</s>

Of course they do.........

If I'm understanding you correctly, this should be called `slug` then (so just remove the `archive`prefix)

Fenn-CS: +1

src/types/sdk/Folder.ts
Fenn-CS: This is the `FacebookRootFolderSpecial = 'facebook.root.folder',` thing we investigated during our pair and concluded it was different from `FolderType`

Which reminds me of something I attempted but removed based on the philosophy in this repo... We have too many special types.


For example:

- `ArchiveStatus`
- `ThumbnailStatus`

and so the need to create more as seen in the PR, you brought up a point about not just leaving these status types as strings because they don't just accept any string BUT we could reduce all these special status types by just having one type `Status` with different strings that all the models use?

src/types/sdk/Folder.ts
Fenn-CS: +1

src/types/sdk/Folder.ts
Fenn-CS: +1

src/types/sdk/Folder.ts
Fenn-CS: I have this uncommitted file in local and this is where I copy the status to create special types, I was just reluctant to suggest this approach, but here we are:

```js


export enum StatusType {
  StatusAccountAllocationAccepted = 'status.account_allocation.accepted',
  StatusAccountAllocationCancelled = 'status.account_allocation.cancelled',
  StatusAccountAllocationExpired = 'status.account_allocation.expired',
  StatusAccountAllocationInProgress = 'status.account_allocation.in_progress',
  StatusAccountSpaceNegative = 'status.account_space.negative',
  StatusAccountSpacePositive = 'status.account_space.positive',
  StatusArchiveGenAvatar = 'status.archive.gen-avatar',
  StatusAuthNeedEmail = 'status.auth.need_email',
  StatusAuthNeedEmailAndPhone = 'status.auth.need_email_and_phone',
  StatusAuthNeedPhone = 'status.auth.need_phone',
  StatusAuthNone = 'status.auth.none',
  StatusAuthOk = 'status.auth.ok',
  StatusAuthUnverified = 'status.auth.unverified',
  StatusAuthVerified = 'status.auth.verified',
  StatusBillingTransferOwnershipAccept = 'status.billing_transfer.ownership_accept',
  StatusBillingTransferOwnershipInProgress = 'status.billing_transfer.ownership_in_progress',
  StatusBillingTransferOwnershipRevoke = 'status.billing_transfer.ownership_revoke',
  StatusConnectorConnected = 'status.connector.connected',
  StatusConnectorDisconnected = 'status.connector.disconnected',
  StatusConnectorNotSetup = 'status.connector.not_setup',
  StatusConnectorUnknown = 'status.connector.unknown',
  StatusContactAccepted = 'status.contact.accepted',
  StatusContactConfirmed = 'status.contact.confirmed',
  StatusContactPending = 'status.contact.pending',
  StatusContactRejected = 'status.contact.rejected',
  StatusContactRevoked = 'status.contact.revoked',
  StatusDirectiveError = 'status.directive.error',
  StatusDirectiveProcessed = 'status.directive.processed',
  StatusDirectiveQueued = 'status.directive.queued',
  StatusDirectiveUnprocessed = 'status.directive.unprocessed',
  StatusFolderCopying = 'status.folder.copying',
  StatusFolderGenthumb = 'status.folder.genthumb',
  StatusFolderMoving = 'status.folder.moving',
  StatusFolderNew = 'status.folder.new',
  StatusGenericDeclined = 'status.generic.declined',
  StatusGenericDeleted = 'status.generic.deleted',
  StatusGenericError = 'status.generic.error',
  StatusGenericInvited = 'status.generic.invited',
  StatusGenericManualReview = 'status.generic.manual_review',
  StatusGenericOk = 'status.generic.ok',
  StatusGenericOrphaned = 'status.generic.orphaned',
  StatusGenericPending = 'status.generic.pending',
  StatusGenericUndefined = 'status.generic.undefined',
  StatusInviteAccepted = 'status.invite.accepted',
  StatusInvitePending = 'status.invite.pending',
  StatusInviteRejected = 'status.invite.rejected',
  StatusInviteRevoked = 'status.invite.revoked',
  StatusNotificationEmailed = 'status.notification.emailed',
  StatusNotificationError = 'status.notification.error',
  StatusNotificationNew = 'status.notification.new',
  StatusNotificationRead = 'status.notification.read',
  StatusNotificationSeen = 'status.notification.seen',
  StatusPromoInvalid = 'status.promo.invalid',
  StatusPromoValid = 'status.promo.valid',
  StatusRecordConverting = 'status.record.converting',
  StatusRecordCopying = 'status.record.copying',
  StatusRecordMoving = 'status.record.moving',
  StatusRecordNeedsConverting = 'status.record.needs_converting',
  StatusRecordNeedsProcessing = 'status.record.needs_processing',
  StatusRecordNeedsThumb = 'status.record.needs_thumb',
  StatusRecordNew = 'status.record.new',
  StatusRecordOnlyMeta = 'status.record.only_meta',
  StatusRecordOverwrite = 'status.record.overwrite',
  StatusRecordProcessing = 'status.record.processing',
  StatusRecordQueued = 'status.record.queued',
  StatusRecordReprocessing = 'status.record.reprocessing',
  StatusRecordUploaded = 'status.record.uploaded',
  StatusRecordUploading = 'status.record.uploading',
  StatusReportError = 'status.report.error',
  StatusReportProcessed = 'status.report.processed',
  StatusReportQueued = 'status.report.queued',
  StatusReportUnprocessed = 'status.report.unprocessed',
  StatusTimezoneDeprecated = 'status.timezone.deprecated',
  StatusTimezoneOk = 'status.timezone.ok',
  StatusZipNew = 'status.zip.new',
  StatusZipOk = 'status.zip.ok',
  StatusZipProcessing = 'status.zip.processing',
  StatusZipRemoved = 'status.zip.removed',
}

```

src/types/sdk/Folder.ts
```diff
@@ -1,3 +1,12 @@
+import type { Item } from './Item';
+
 export interface Folder {
+  id: number;
+  name: string;
+  slug: string;
```
slifty, line 6: This is going to be populated by `archiveNbr` which isn't current in the FolderVo definition, so that should be added.

src/types/sdk/Folder.ts
```diff
@@ -1,3 +1,12 @@
...
   size: number;
+  displayDate: Date;
+  readonly children: (Folder | Item)[];
+  readonly createdAt: Date;
```
slifty, line 10: These need corresponding entries in the `FolderVo`

src/types/sdk/Folder.ts
```diff
@@ -1,3 +1,12 @@
+import type { Item } from './Item';
+
 export interface Folder {
+  id: number;
+  name: string;
+  slug: string;
   size: number;
+  displayDate: Date;
+  readonly children: (Folder | Item)[];
```
slifty, line 9: This should be two separate fields:

`folders` and `items` <-- this allows end users to access without having to explicitly type guard / do some other form of type check -- they can simply know that `folders` contain `Folder` objects etc.

src/utils/folderVoToFolder.ts
```diff
@@ -0,0 +1,22 @@
+import type {
+  FolderVo,
+  Folder,
+} from '../types';
+
+export const folderVoToFolder = (folderVo: FolderVo): Folder => {
+  const createdAt = new Date(folderVo.createdDT); 
+  const updatedAt = new Date(folderVo.updatedDT);
+  const displayDate = new Date(folderVo.displayDT);
+  const slug = `${folderVo.archiveNbr}/${folderVo.folderId}`;
```
slifty, line 10: Argh -- slug feels incorrect here (this is really TWO slugs -- one is the archive slug the other is the folder slug) being combined into a single item.

What if we had `slug` and `archiveSlug` as two separate fields?  That would be more accurate naming.

Alternatively we could do the concat and call it something else, e.g. something like `permanentUriPath` or `relativeUriPath`

That said, I don't think I need this for sftp at all so we could even consider just... ditching this field entirely and punting the decision.  Thoughts?

Fenn-CS: Yes, especially because we have no use for it now and could construct if we know what we want. I would submit and update that removes the slug from both types.

Fenn-CS: Removed.

src/types/api/FolderVo.ts
```diff
@@ -1,3 +1,21 @@
-// This is a placeholder interface which will ultimately be populated
-// eslint-disable-next-line @typescript-eslint/no-empty-interface
-export interface FolderVo {}
+export interface FolderVo {
+  // These fields map to those defined in the base library ArchiveVO
```
slifty, line 2: Should this be `FolderVO`

src/types/api/FolderVo.ts
```diff
@@ -1,3 +1,21 @@
...
+  // https://github.com/PermanentOrg/back-end/blob/main/api/core/folder/vo/folder.vo.php
+  folderId: number;
+  archiveId: number;
```
slifty, line 5: I don't think we use this any more now that we took out slug.  Let's take it out.

src/types/api/FolderVo.ts
```diff
@@ -1,3 +1,21 @@
...
+  archiveNbr: number;
```
slifty, line 6: Same here -- I don't think we're using archiveNbr any longer.

slifty (changes requested):
Tiny tweaks -- Let's take out anything in FolderVo that isn't directly used when mapping to a real Folder, since it's just a dead attribute for now, unless it's used in that conversion.

(I imagine once we'll add that in later, but for *now* it isn't used)

src/types/api/FolderVo.ts
Fenn-CS: @slifty just to be clear since 3 lines are highlighted... I am taking out `archiveId` and `archiveNbr` right?

This should also happen in #31 which records I presume?

src/types/api/FolderVo.ts
Fenn-CS: `FolderVO` is what we used in the back-end however for this repository I just maintained the notation I found you using earlier. I think either of them is fine here. What we are using here is actual camel case.

src/types/api/FolderVo.ts
Fenn-CS: Taken out.

Fenn-CS:
Took out `archiveId` and `archiveNbr` because we would definitely used the `folderId`. (Mentioning this because it was marked in your review).

So please do check that, that is what was expect and resolve comments.

Thank you!

cc: @slifty

src/types/api/FolderVo.ts
slifty: I'm sorry I wasn't clear -- I meant 

```
  // These fields map to those defined in the base library ArchiveVO
```

I think that should be `in the base library FolderVO`

Fully agree `FolderVo` is the right casing.

src/types/api/FolderVo.ts
slifty: Correct!

slifty: (This is ready to resolve but didn't want you to miss the github UX comment since it might be useful knowledge in future)

src/types/api/FolderVo.ts
Fenn-CS: Gotcha! Thanks again!

slifty (approved):
Looks great thank you!

Fenn-CS:
Thanks @slifty

PR #27, opened 2022-05-09
Downgrade node-fetch to v2

### LINKED ISSUE #25: node-fetch v3+ doesn't support common JS
The latest version of `node-fetch` [does not support CJS, only ESM](https://github.com/node-fetch/node-fetch#commonjs).

This is a problem for us because, at this point anyway, we want to support CJS.  The result is that we have to downgrade node fetch to v2, otherwise our CJS build breaks in any project that uses it with an error along the lines of:

```
/Users/slifty/Maestral/Code/open-tech-strategies/permanent/permanent-sdk/dist/cjs/utils/makePermanentApiCall.js:7
const node_fetch_1 = __importDefault(require("node-fetch"));
                                     ^

Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/slifty/Maestral/Code/open-tech-strategies/permanent/permanent-sdk/node_modules/node-fetch/src/index.js from /Users/slifty/Maestral/Code/open-tech-strategies/permanent/permanent-sdk/dist/cjs/utils/makePermanentApiCall.js not supported.
Instead change the require of index.js in /Users/slifty/Maestral/Code/open-tech-strategies/permanent/permanent-sdk/dist/cjs/utils/makePermanentApiCall.js to a dynamic import() which is available in all CommonJS modules.
    at Object.<anonymous> (/Users/slifty/Maestral/Code/open-tech-strategies/permanent/permanent-sdk/dist/cjs/utils/makePermanentApiCall.js:7:38)
```

slifty:
This PR downgrades node-fetch to v2 from v3

The API is the same for our purposes, it's just that node-fetch v3 doesn't support CJS and we need it to.

I didn't add a test around this even though it's a fix to a bug because I'm not really sure how / it is more of an integration issue and we don't have integration tests.  I only learned about it by importing the compiled package into a CJS project and seeing that node-fetch was unhappy.)

Resolves #25

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,12 +9,13 @@
       "version": "0.0.0",
       "license": "MIT",
       "dependencies": {
-        "node-fetch": "^3.2.0"
+        "node-fetch": "^2.6.7"
       },
       "devDependencies": {
         "@tsconfig/esm": "^1.0.0",
         "@tsconfig/node14": "^1.0.1",
         "@types/jest": "^27.4.0",
+        "@types/node-fetch": "^2.6.1",
         "@typescript-eslint/eslint-plugin": "^5.10.0",
         "@typescript-eslint/parser": "^5.10.0",
         "eslint": "^8.7.0",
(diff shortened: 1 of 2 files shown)


Fenn-CS (approved):
Looks good.

Another note: Why do we need CJS here? AFAIK, I see consistent use of ES16 module and importing style here.

Nonetheless, it might be useful to abandon CJS. It's only a matter of time before this becomes a thing with a bunch of dependencies.

slifty:
> Why do we need CJS here?

So the lack of support for CJS is cascading -- meaning we can't even transpile to CJS while using node-fetch v3 without producing a broken build.  We want to transpile to CJS because we want users who rely on / transpile to CJS to be able to use our SDK.

Since this is a library we're publishing we are limited if our dependencies don't support our targets.

jasonaowen (approved):
+1 to this; ESM-only modules are very frustrating, and for this particular dependency, our long-term goal is to use the native `fetch` once all supported Node versions have that available.

PR #28, opened 2022-05-09
Add getArchives SDK method

### LINKED ISSUE #4: Implement archive access

slifty:
This PR is adds our very first piece of actual functionality to the SDK: the ability to get all archives for an authenticated account.

It also:

1. Introduces [ajv](https://www.npmjs.com/package/ajv) which is a nice runtime counterpart to TypeScript that lets you define schemas that can be easily turned into type guards.

2. Adds the concept of a `ClientConfiguration` which will allow our SDK users to provide not just their authentication credentials to the SDK but also anything else needed to form their "client" (for now the only extra field is the permanent base url which is needed to allow us to interact with other instances of the permanent API aside from prod)

3. Adds `nock` which is a tool we can use to write unit tests that rely on network calls.

Resolves #4


slifty:
I want to add tests so we have 100% coverage before this is merged, but I do think it is ready for at least a first pass review @Fenn-CS

slifty:
In addition to tests I think we should at least document *something*.  I don't want to have to figure out the final pattern for documentation, but even an "DOCUMENTATION.md" or something would be a fine start.

I also need to update the changelog before merging.

slifty:
This commit has way too much going on; I can split it into more than one if that would be helpful.

cecilia-donnelly:
I think it's a good idea to have @meisekimiu look at this one, as well.

slifty:
note that the drop in test coverage is because we added a bunch of untested files to the pool of files that are considered by the test coverage report.

slifty:
I bent to my own pressure and opened up this PR for the opening "grab bag" commits https://github.com/PermanentOrg/permanent-sdk/pull/36

Once that's merged I'll rebase (and, for instance, the code coverage report will actually be meaningful)

Fenn-CS (changes requested):
I think that it would be nice if we can break down this PR to multiple ones. Things I was able to see quickly:

- Node-fetch, not an issue but `axios` comes with more benefits like automatic To-JSON conversion right? It would be nice to know if you have any specific reasons for going with `node-fetch` (Don't know much about it yet)

- The naming of some files and functions for example `makePermanentAPICalls` and `formatTimestampAsUtc` implies the kind of naming scheme where every function would probably have its own file which might lead us to having too many files in the long run.

I think that a file name like `timeUtils` could contain `formatTimestampAsUtc` and then we can add any other time utils as functions into that file in future. Same thing with `makePermenentAPICalls` we probably want define a minimum `http` file with basic http helpers. 

- I would suggest adding a `services` folder and we can add `permanetService.ts` that would reference the `http` file above, that way the rest of the code can use the `permanentService` (perhaps a class to interact with the rest of the API`

src/utils/__tests__/formatTimestampAsUtc.test.ts
```diff
@@ -0,0 +1,12 @@
+import { formatTimestampAsUtc } from '..';
+
+describe('formatTimestampAsUtc', () => {
+  it('should not modify a UTC timestamp', () => {
+    const timestamp = '2022-07-10T14:59:03.853Z';
+    expect(formatTimestampAsUtc(timestamp)).toEqual(timestamp);
+  });
+  it('should not add Z to a non-UTC timestamp', () => {
```
meisekimiu, line 8:
```suggestion
  it('should add Z to a non-UTC timestamp', () => {
```
Tiny nitpick, but the "should not" seems like it should just be "should" right?

slifty: Great catch!  Fixing.

Fenn-CS:
> Either way, the "one utility per file" IS the pattern we have today; what do you think of creating an issue for that refactoring where we can discuss what's best at that time instead of trying to overhaul utility organization in this PR?

That's reasonable. I should mention that I started "breaking that pattern" in https://github.com/PermanentOrg/permanent-sdk/blob/main/src/utils/generateEnumTypeguard.ts where instead of adding an extra `isRecordStatus.ts` just like with the now deleted [isThumbnailStatus](https://github.com/PermanentOrg/permanent-sdk/commit/bbf5e7d7ae76ad1cc48c83748b36522eb6a6fb99#diff-3fd8b0fd57e8c162d8d0414ed15aef21ee36a9aeb2bf4a833d5260a2969453d7) I threw all of that into `generateEnumTypeguard.ts` perhaps it may be worth while renaming such a file to reflect or filing a PR get things back in line with the correct way of doing things. 

I would open an issue for that discussion. 

What do you think about the service class idea here though?

slifty:
I may have misunderstood the idea of the service class -- right now we have full functionality without a class (and I believe we are trying to avoid using classes in this repository in general).

What would a class give us exactly that isn't already covered?

I figured each sdk call with have its own function that invokes `makePermanentApiCall` -- so that's the only http helper we're ever going to need.

slifty:
Just to clarify the intended design here:

We have `/src/api/getAllArchiveVos.ts` which interacts with the endpoint to get all archives

We might then have things like:

* `/src/api/getRecordVos.ts`
* `/src/api/getFolderVos.ts`
* `/src/api/deleteRecordVo.ts`

^ so basically the `/src/api` is the "class" (but in a functional / tree shake-able structure so an application that never uses `deleteRecordVo` wouldn't have to include that file in its bundle)

Does that design approach make sense?

Fenn-CS:
> I figured each sdk call with have its own function that invokes makePermanentApiCall, passing in the endpoint in question -- so that is the only http helper we're ever going to need.
> 
> What would a class give us that isn't already covered by that function?

I already see that we have files such as `getArchiveVo.ts` in the class model all of those would methods in the `permanentService`  class (which we could easily let go if there no other API this would ever need to interact without passing through the backend) BUT all of this still boils down the code organization issue which we should probably discuss on its own ticket as you have pointed out.

I was just mentioning because the style guide for thing kind of thing is only implicitly written in the code base itself and with how small the SDK is at the moment these kind of discussions could not easily come up. So it's a good thing to have in mind.

If this meets me,  should be able to do a detailed code review tomorrow!

slifty:
Ah understood thank you for that explanation; and totally appreciate raising it now, since as you say we aren't too invested yet.

I do think that we should stick with the functional approach -- some of the motivation is the same as the points that came out of our larger "classes vs functional" discussion early on when we landed on functional: pure functions are self documenting, and tree shaking is a vital feature in an sdk (Classes cannot be tree shaken; functions can).

Re the other services question -- I don't think we'll need to work against other APIs BUT if we do I think we could take that opportunity to just shift the file structure a bit, e.g. rename `/src/api` to `src/services/permanent` and keep the functions otherwise the same.

src/utils/__tests__/makePermanentApiCall.test.ts
```diff
@@ -0,0 +1,58 @@
...
+    const response = await makePermanentApiCall(
+      {
+        bearerToken: '12345',
+        baseUrl: 'https://permanent.local/api',
+      },
+      '/testing',
+    );
+    const responseBody = await response.text();
+
+    expect(responseBody).toEqual('it worked');
```
Fenn-CS, line 33: Looks like `node-fetch` does not return the `request` param in the `Response` object...  If it did, it would be nice addition to have another assertion here that looks like : `expect(response.request.url).toEqual('https://permanent.local/api');`

Fenn-CS (approved):
Nice pace setter for how we would continue to make calls to back-end like the case of `Folders` and `Records.`

Looks good, left some two minor comments about an additional assertion that could make one test clearer if it was available and about the introduction of libraries with their use.

Thanks @slifty

src/utils/__tests__/makePermanentApiCall.test.ts
Fenn-CS: Awesome!

package.json
```diff
@@ -56,10 +56,12 @@
     "eslint-plugin-import": "^2.25.4",
     "eslint-plugin-sort-exports": "^0.6.0",
     "jest": "^27.4.7",
+    "nock": "^13.2.8",
```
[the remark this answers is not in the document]
slifty: That sounds great to me

PR #29, opened 2022-05-23
Add link:fix command

Fenn-CS:
Runing npm run lint:fix would now easily repair the errors
that are shown by npm run lint.

Signed-off-by: fenn-cs <fenn25.fn@gmail.com>


jasonaowen (approved):
Yeah, that seems helpful!

Fenn-CS:
Thanks @jasonaowen

PR #31, opened 2022-05-24
Add API & SDK type defs for RecordVO

### LINKED ISSUE #30: Create record types + utilities
Similar to #18  -- we need to create the Item and RecordVo types, plus create the recordVoToRecord utility.

Fenn-CS:
This PR add the base required properties for RecordVO from back-end just like in #24 and #26.

Resolves #30


jasonaowen (changes requested):
Can you please also include the code that converts between a `Record` and a `RecordVO`? Without that code, it's really hard to judge these models.

src/types/sdk/Record.ts
```diff
@@ -0,0 +1,14 @@
+import type { Status } from './Status';
+import type { RecordType } from './RecordType';
+
+export interface Record {
+  id: number;
+  slug: string;
```
slifty, line 6: I think this is the *archive* slug, not the record slug, so let's call it `archiveSlug` to be more explicit.

src/types/sdk/RecordType.ts
```diff
@@ -0,0 +1,20 @@
...
+  Gedzip = 'Genealogy Archive',
+  Image = 'type.record.image',
+  PDF = 'type.record.pdf',
+  Presentation = 'type.record.presentation',
+  Reference = 'type.record.reference',
+  Spreadsheet = 'type.record.spreadsheet',
+  Unknown = 'type.record.unknown',
+  Video = 'type.record.video',
+}
+
```
slifty, line 19: Nit pick but we can toss this white space

slifty (changes requested):
The big thing we should add to this is the `FileVo` data.

slifty (commented):
Github UX is completely glitching out and marking my comments as part of a full review so. here's this.

src/types/sdk/Record.ts
jasonaowen: I think @meisekimiu knows more about this than I do, so I'll defer to her on how URLs are (de)constructed.

However, it is my (incomplete, untested) understanding that each archive, record, and folder have a globally unique `archiveNbr`. It may be the case that web-app URLs and share links have multiple `archiveNbr`s; I haven't looked into that.

src/types/api/RecordVo.ts
```diff
@@ -0,0 +1,22 @@
...
+  // These fields map to those defined in the base library ArchiveVO
+  // https://github.com/PermanentOrg/back-end/blob/main/api/core/record/vo/record.vo.php
+  recordId: number;
+  archiveId: number;
+  type: string;
+  status: string;
+  displayName: string;
+  uploadFileName: string;
+  displayDT: string;
+
```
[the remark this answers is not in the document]
Fenn-CS: This has been added.

Fenn-CS:
@slifty I added a commit to redefine the `BaseVo` the details are in that particular commit. I used use the `BaseVo` as the basis for all the `API` types because that's out it works on back-end and would avoid inconsistency errors. Like the `updateDT` versus `updatedDT` which I sighted in the commit message.


Thanks!

src/types/sdk/Record.ts
```diff
@@ -0,0 +1,14 @@
+import type { Status } from './Status';
+import type { RecordType } from './RecordType';
+
+export interface Record {
+  id: number;
+  displayDate: Date;
+  type: RecordType;
+  name: string;
+  fileURL: string;
+  downloadURL: string;
```
slifty, line 10: Let's do `Url` instead of `URL` since I believe it's consistent with our other naming decisions (one factor here is we wouldn't say `uRLValidator` we would say `urlValidator`)

This applies to both `fileUrl` and `downloadUrl`

src/types/sdk/RecordType.ts
```diff
@@ -0,0 +1,18 @@
...
+export enum RecordType {
+  Archive = 'type.record.archive',
+  Audio = 'type.record.audio',
+  Document = 'type.record.document',
+  Experience = 'type.record.experience',
+  Folder = 'type.record.folder',
+  Gedcom = 'Genealogy',
+  Gedzip = 'Genealogy Archive',
+  Image = 'type.record.image',
+  PDF = 'type.record.pdf',
```
slifty, line 12: Similar to `URL` => `Url` let's do `Pdf` here

src/types/sdk/RecordType.ts
```diff
@@ -0,0 +1,18 @@
+// These types are defined in the permanent backend:
+// https://github.com/PermanentOrg/back-end/blob/main/library/definition/perm.constants.php
+export enum RecordType {
+  Archive = 'type.record.archive',
+  Audio = 'type.record.audio',
+  Document = 'type.record.document',
+  Experience = 'type.record.experience',
+  Folder = 'type.record.folder',
+  Gedcom = 'Genealogy',
+  Gedzip = 'Genealogy Archive',
```
slifty, line 10: Just a sanity check: is Gedzip one word or is it meant to be a compound word (e.g. `GedZip`)

slifty (approved):
This looks great!

Some very minor capitalization tweaks I think would be good for consistency, but other than that go forth!  I'm marking as approved because I don't think this will need a re-review after those casing changes.

I like the use of `BaseVo` good thinking.

src/types/sdk/RecordType.ts
Fenn-CS: Recommendations applied.

Fenn-CS:
Thanks for the review @slifty , the capitalization changes where applied accordingly!

PR #41, opened 2022-07-14
Allow manual dispatch of release flow

slifty:
This PR makes it possible to manually trigger the `release` workflow.


cecilia-donnelly (approved):
:+1:

PR #44, opened 2022-07-14
Specify public package

slifty:
This PR tells npm that we want to publish the package publicly.

Without this, npm rejects the package since private packages are a paid features (plus we don't want it private anyway so)


cecilia-donnelly (approved):
I am learning so much about npm this afternoon!

PR #45, opened 2022-07-18
Support "archive folder" access

slifty:
This PR adds a new SDK function: `getArchiveFolders` which will return the immediate child folders of an archive.

For the user of an SDK this means they don't have to think about the concept of a "root folder" for an archive but instead can just directly load the system folders for a given (e.g. "My Files" or "Apps") using this function.

This builds on #46 so probably worth waiting to review until that is merged.


slifty:
@Fenn-CS I think this is now ready for review whenever you get a chance!

src/types/api/FolderVo.ts
```diff
@@ -24,6 +28,17 @@ export const folderVoSchema: JSONSchemaType<FolderVo> = {
...
       type: 'integer',
     },
+    ChildItemVOs: {
+      type: 'array',
+      items: {
+        type: 'object',
+        required: [],
+        anyOf: [
+          { $ref: '#' },
+          recordVoSchema,
```
Fenn-CS, line 38: Shouldn't we have the `folderVoScheme` here as well?

Fenn-CS (approved):
I left one inline comment about adding the `folderVoSchema` to the validation of `childItemVo`s otherwise this looks good.

src/types/api/FolderVo.ts
slifty: So you are absolutely right that the child items can be a record OR a folder; in json schema the way you write a self-referencing schema is by `{ $ref: '#' }` so that's what that first line is doing (the schema being defined is, itself, the folder schema so I can't reference it by object).

slifty:
Thanks @Fenn-CS !  I think that my answer addresses the concern (folder is part of the validation so we should be all set).

I'll give a few moments before merging in case you have a followup concern but if not I think this is good to go.

src/sdk/getArchiveFolders.ts
```diff
@@ -0,0 +1,19 @@
...
+  Archive,
+  ClientConfiguration,
+  Folder,
+} from '../types';
+import { getArchiveRootFolderVo } from '../api';
+import { folderVoToFolder } from '../utils';
+
+export async function getArchiveFolders(
+  clientConfiguration: ClientConfiguration,
+  archive: Pick<Archive, 'id'>,
```
slifty, line 11: @meisekimiu @Fenn-CS what do you think of this pattern.

Should it take an `archive` that only is expected to have an id, or should I have the dev user pass just an `archiveId`.

The difference would be:

```
clientConfig = { ... };
archive = sdk.getArchives(clientConfig)[0];
folders = sdk.getArchiveFolders(
  clientConfig,
  archive,
);
```

vs

```
clientConfig = { ... };
archive = sdk.getArchives(clientConfig)[0];
folders = sdk.getArchiveFolders(
  clientConfig,
  archive.id,
);
```

(Apply this pattern to other parameters)

src/types/api/FolderVo.ts
Fenn-CS: Thanks for the explanation!

src/sdk/getArchiveFolders.ts
Fenn-CS: Using the type Archive with ID provides some consistency but can sometimes add extra steps such as creating the type and intializing it with an id. If we won't be doing such intermediate preparations to call the get methods this is fine!

Otherwise I would lean more towards just passing their `ids`

Fenn-CS: Using the type Archive with ID provides some consistency but can sometimes add extra steps such as creating the type and intializing it with an id. If we won't be doing such intermediate preparations to call the get methods this is fine!

Otherwise I would lean more towards just passing their `ids`

slifty: I'm gonna modify this to just pass the Ids!  This pattern becomes even more clearly correct when we have situations where we need to pass both a folderId and an archiveId (which we will have to do later for other calls unfortunately)

PR #49, opened 2022-07-21
Use arrow functions

slifty:
This PR fixes two function definitions that accidentally wasn't using arrow functions.

I don't see a great way to check this via linting which is frustrating.  Maybe some day.


Fenn-CS (approved):
Looks good, thanks! =====>

PR #51, opened 2022-07-26
Add the File type

### LINKED ISSUE #50: We should expose `File` to the sdk user
It turns out that FileVos serve an important purpose, since a record can have more than one file derivative in the system.  We should expose those derivatives to SDK users to allow them to determine which version of the record they want to work with.

This means adding `File` to our `Record` type (note that this impacts #47 as we think about better names for `Record`)

slifty:
This PR creates the concept of a `File` type, which lets records have more than one associated file for SDK users.

Resolves #50


src/types/sdk/FileFormat.ts
```diff
@@ -0,0 +1,12 @@
+import { generateEnumTypeguard } from '../../utils/generateEnumTypeguard';
```
Fenn-CS, line 1: The name of this file sounds like it's referring to extensions which represents formats actually it appears to be reffering to `CoversionStatus` or maybe `FileConversionStatus`.

Fenn-CS (commented):
Have a minor concern.

src/types/sdk/FileFormat.ts
slifty: That's a fair concern.  Worth noting that `format` is the internal name, and so there is a subtle cost from any type deviation.  That said, I could appreciate that `format` is a generic term that might be worth getting rid of here.

I think it's not a conversion status (status indicates something that will change to me).  In the world of Archive.org this would be a `derivative type` (each of these files are a different kind of derivative of the original).

Maybe:

* `DerivativeType`
* `VersionType`

I have to keep thinking on this one....

Fenn-CS: @slifty It would be nice to have it be `DerivativeType` that's way clearer. I think it's okay if we change it because it's happening in the `sdk` namespace if we had or have to define it in the `api` namespace we can maintain the naming scheme from back-end and that conversions would show in the utility that would convert that.

Umm, it's a minor concern still if you strongly prefer to keep this way then I could pass on it.

slifty: The concern is very valid!

Your strong response to `DerivativeType` sells it for me!  I'll make that change.

Fenn-CS (approved):
LGTM! Thanks!

PR #58, opened 2022-08-02
Add getRecord to the SDK

### LINKED ISSUE #42: Support record access
The SDK needs to support the ability to access records associated with a folder.

slifty:
This PR adds a `getRecord` function to the sdk's API, allowing SDK users to load records from Permanent.

Resolves #42


src/api/getRecordVo.ts
```diff
@@ -0,0 +1,35 @@
...
+  recordId: number,
+  archiveId: number,
+): Promise<RecordVo> => {
+  const queryParams = new URLSearchParams({
+    recordId: `${recordId}`,
+    archiveId: `${archiveId}`,
+  });
+  const response = await makePermanentApiCall(
+    clientConfiguration,
+    `/record/getById?${queryParams.toString()}`,
```
Fenn-CS, line 31: The endpoint that supports `version 2` is `/api/record/get` the call structure remains the same.

Fenn-CS (commented):
I am not sure if you tested this against a branch locally but please clerify `api/get/getById` does not support version 2 requests yet.

Otherwise this looks great and I would wait for clarification.

src/api/getRecordVo.ts
Fenn-CS: Essentially, the endpoint that we have started working on to support `v2` is the one I stated above. 

See : https://github.com/PermanentOrg/back-end/pull/252

slifty: Ahh thank you great catch!  I had the wrong endpoint.

Fenn-CS (approved):
Great! Thanks @slifty !

src/api/getRecordVo.ts
Fenn-CS: Cool, now looks good and I would mar this as resolved.

slifty:
I added a few little things to this PR because I found some additional fields that were missing in the current File / Record which are needed for the rclone project.  Specifically: `size` and `downloadName`

PR #63, opened 2022-12-16
Add `fileSystemCompatibleName` attributes

### LINKED ISSUE #62: Better support for `downloadName` / `fileSystemName`
The server side has the concept of "downloadName" for folders and records (and archives? need to check).

We translate this to an attribute of our Record, but it needs to also exist for other entities.  I also think that from the SDK point of view `fileSystemName` would be a more descriptive term for the attribute (`download` indicates a more specific, though clearly related, use case!)

slifty:
The `fileSystemCompatibleName` is a version of a given record or folder's name that will work correctly on a file system.  This means it will not have illegal characters and will not conflict with peer items.

This is a breaking change for any client that was using the `Record.fileName` attribute.

Resolve #62


src/types/sdk/Folder.ts
```diff
@@ -5,6 +5,7 @@ export interface Folder {
   name: string;
   size: number;
   displayDate: Date;
+  readonly fileSystemName: string;
```
Fenn-CS, line 8: `fileSystemName` sounds like it's referring to something else!

`nameOnFileSystem`, `fileSystemCompatibleName` longer but clearer names.

I won't die on this hill though!

Fenn-CS (commented):
Overall, this looks good function-wise. I have a minor reservation about the name `fileSystemName` to which I left an inline comment.

Lastly, what is the motivation for not simply using the existing `fileName` property that existed already and correctly mapped `downloadName` (which to the best of knowlege is file system compatible).

src/types/sdk/Folder.ts
slifty: Oh right -- that sounds like it's referring to the name of the file system!  No need to die on the hill I'll join you there and try to identify a better solution :D

slifty: Using `fileSystemCompatibleName`

Also to answer the question below; `fileName` wasn't very clear and also didn't quite make sense as a name on the `Folder` type, so I think this is an improvement.

Fenn-CS (approved):
Looks good!

PR #67, opened 2023-01-03
Rename Record to ArchiveRecord

### LINKED ISSUE #47: Oh no, Record is already a thing
TypeScript has a built in concept of [Record](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) built into the language.  Unfortunately the Permanent backend uses the term `Record` to represent a file in the permanent system.

For the API interaction we're OK because it's actually not `Record` but `RecordVo`.  However, we're going to need to think of something else for our sdk side that isn't just `Record`.

Some ideas:

1. `PermanentRecord` <-- ugh.
2. `File` (this could be confusing since we have `FileVo` which is... different than `RecordVo` somehow)
3. `Item` <-- this has the same downside of `File` while also being very generic.
4. `Document`

Nothing's perfect.  I think `File` might be our best bet (as a partner to `Folder`).  I'm 99% sure that we're merging `RecordVo` and `FileVo` anyway; and if we need to keep track of whatever `FileVo` means we can think of a more specific name for the sdk (e.g. `FileInstance` or `FileVersion` or something).

Any thoughts?

slifty:
This PR renames `Record` to `ArchiveRecord` in order to avoid a name collision with the [TypeScript utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type)

Resolves #47


slifty:
There was some discussion today about changing this to `ArchiveRecord` (I may shift this to `ArchivalRecord` to avoid any sense that this is directly related to our concept of a metadata associated with an `Archive`) instead of `Item` for a few reasons:

1) Item is already a term in the API (that means either a "record" or "folder") <-- this is not the sole reason, since that terminology may need to disappear anyway
2) ArchivalRecord is a term of art that is probably why `Record` was used in the first place, so it's reasonable to stick with it.

src/sdk/getArchiveRecord.ts
```diff
@@ -1,18 +1,18 @@
...
 } from '../types';
 import { getRecordVo } from '../api';
-import { recordVoToRecord } from '../utils';
+import { recordVoToArchiveRecord } from '../utils';
 
-export const getRecord = async (
+export const getArchiveRecord = async (
   clientConfiguration: ClientConfiguration,
-  recordId: number,
+  itemId: number,
```
slifty, line 10: This should be archiveRecordId

src/utils/recordVoToArchiveRecord.ts
```diff
@@ -1,30 +1,30 @@
...
 );
 
-export const recordVoToRecord = (recordVo: RecordVo): Record => {
+export const recordVoToArchiveRecord = (recordVo: RecordVo): ArchiveRecord => {
   const createdAt = new Date(formatTimestampAsUtc(recordVo.createdDT));
   const updatedAt = new Date(formatTimestampAsUtc(recordVo.updatedDT));
   const displayDate = new Date(formatTimestampAsUtc(recordVo.displayDT));
-  const type = recordTypeToRecordType(recordVo.type);
-  const status = recordStatusToStatus(recordVo.status);
+  const type = recordVoTypeToItemType(recordVo.type);
```
Fenn-CS, line 26: Is this not supposed to be `recordVoTypeToArchiveRecordType` ?

src/utils/recordVoToArchiveRecord.ts
```diff
@@ -1,30 +1,30 @@
...
-  isRecordType,
+  isArchiveRecordType,
   isStatus,
 } from '../types';
 import { fileVoToFile } from './fileVoToFile';
 import { formatTimestampAsUtc } from './formatTimestampAsUtc';
 
-const recordTypeToRecordType = (recordType: string): RecordType => (
-  isRecordType(recordType) ? recordType : RecordType.Unknown
+const recordVoTypeToItemType = (recordType: string): ArchiveRecordType => (
```
Fenn-CS, line 14: Why `ItemType` suddenly and not `ArchiveRecordType`?

slifty: Thank you!  Fixed :D

Fenn-CS (approved):
Thanks! Looks good now!

PR #70, opened 2023-01-20
Noissue organization changes

slifty:
This PR makes some minor changes to the code organization.

None of these changes should have any impact on the SDK functionality or exposed API.

### DIFF
--- a/src/api/createFolderVo.ts
+++ b/src/api/createFolderVo.ts
@@ -34,5 +34,6 @@ export const createFolderVo = async (
       body,
     },
   );
-  return parseFolderVo(await response.json());
+  const responseText = await response.text()
+  return parseFolderVo(JSON.parse(responseText));
 };
--- a/src/api/createFolderVo.ts
+++ b/src/api/createFolderVo.ts
@@ -1,21 +1,13 @@
-import { ValidationError } from '../errors';
 import { isFolderVo } from '../types';
-import { makePermanentApiCall } from '../utils';
+import {
+  makePermanentApiCall,
+  typedJsonParse,
+} from '../utils';
 import type {
   ClientConfiguration,
   FolderVo,
   Folder,
 } from '../types';
-
-const parseFolderVo = (value: unknown): FolderVo => {
-  if (isFolderVo(value)) {
-    return value;
-  }
-  throw new ValidationError(
-    'Invalid server response format',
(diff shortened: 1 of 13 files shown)


src/utils/__tests__/typedJsonParse.test.ts
```diff
@@ -0,0 +1,57 @@
...
+      foo: 'hello',
+      bar: {
+        baz: 42,
+      },
+    };
+    const jsonString = JSON.stringify(target);
+    expect(typedJsonParse(jsonString, isTarget)).toEqual(target);
+  });
+  it('should throw an error when provided a JSON string with an unexpected shape', () => {
+    const miss = {
```
Fenn-CS, line 51: I am surprised that this object does not get converted to valid json. Out of curiosity what is wrong here?

src/api/createFolderVo.ts
```diff
@@ -34,5 +34,6 @@ export const createFolderVo = async (
       body,
     },
   );
-  return parseFolderVo(await response.json());
+  const responseText = await response.text()
+  return parseFolderVo(JSON.parse(responseText));
```
Fenn-CS: This question applies to everything concerning the response processing commit.

Nothing wrong, just a curious question, the various parse methods like `parseFolderVo` don't fail when server returns invalid json?

`JSON.parse` would fail in the current structure and output the text that hints on what is wrong?

Fenn-CS (approved):
+1 on the default object removal! It had become and unwritten rule meanwhile we don't even use it.

I left some questions for my own use inline. Thanks!

src/utils/__tests__/typedJsonParse.test.ts
slifty: It *is* valid JSON but it is *not* the right shape (it doesn't result in a `Target` typed object).

Basically this method takes a string AND a type guard and it:

1) parses the JSON string
2) then compares the resulting object with the type guard and errors if it is not the correct type.

Fenn-CS: True, thanks! I over generalized the test.

src/api/createFolderVo.ts
Fenn-CS: Thanks Dan!

PR #79, opened 2023-03-02
Add slug to Archive type

slifty:
This PR creates the concept of a `slug` for the `Archive` type.

Note that `archiveNbr` is a legacy name and is a bit confusing since it is not actually a number; this PR is proposing `slug` as a more meaningful term to refer to the attribute.  Slug is an existing concept: https://developer.mozilla.org/en-US/docs/Glossary/Slug

Resolves #87

### DIFF
--- a/src/utils/__tests__/archiveVoToArchive.test.ts
+++ b/src/utils/__tests__/archiveVoToArchive.test.ts
@@ -4,6 +4,7 @@ describe('archiveVoToArchive', () => {
   it('should properly populate an Archive based on ArchiveVo data', () => {
     const archiveVo = {
       archiveId: 0,
+      archiveNbr: '0000-0000',
       fullName: 'archive name',
       createdDT: '2022-07-10T14:59:03.853Z',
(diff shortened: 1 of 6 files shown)


src/utils/__tests__/archiveVoToArchive.test.ts
```diff
@@ -4,13 +4,15 @@ describe('archiveVoToArchive', () => {
...
       archiveId: 0,
+      archiveNbr: '000-000',
       fullName: 'archive name',
       createdDT: '2022-07-10T14:59:03.853Z',
       updatedDT: '2022-08-10T14:59:03.853Z',
     };
     const archive = archiveVoToArchive(archiveVo);
     expect(archive).toEqual({
       id: 0,
+      slug: '000-000',
```
cecilia-donnelly, line 15: I suppose it doesn't matter, but in practice I believe these are always 4 characters on either side of the hyphen.

cecilia-donnelly (approved):
Thank you!

PR #84, opened 2023-03-29
Add `deleteFolder`

### LINKED ISSUE #34: Support folder deletion
The SDK needs to support the deletion of existing folders in an archive.  This might be through a `deleteFolder` function.

There is some discussion of disallowing SFTP deletion within certain archives; I think that will be an SFTP implementation (and that the SDK wouldn't have that restriction; it would be the tools implementing the SDK that should pay attention to those settings)

slifty:
This PR adds the ability to delete folders via the sdk.

Resolves #34


Fenn-CS (approved):
LGTM.

meisekimiu (approved):
Also adding an approval to this! :+1:

PR #85, opened 2023-03-29
Add `deleteArchiveRecord` to sdk

### LINKED ISSUE #35: Support record deletion
Similar to #34 The SDK needs to support the deletion of existing records in an archive. This might be through a `deleteRecord` function.

slifty:
This PR makes it possible to delete records on permanent via sdk.

Resolves #35


Fenn-CS (approved):
Looks great!

meisekimiu (approved):

PR #87, opened 2023-03-30
Add getAuthenticatedAccount

### LINKED ISSUE #86: Support the ability to load account preferences
For the sftp service we want to create a "safety valve" setting that disallows deletion via sftp unless the account has explicitly enabled it.  Clients need to be able to access this type of setting via the sdk.

slifty:
This PR adds the `getAuthenticatedAccount` function to the sdk.

Resolves #86


src/utils/accountVoToAccount.ts
```diff
@@ -0,0 +1,12 @@
+import type {
+  AccountVo,
+  Account,
+} from '../types';
+import { formatTimestampAsUtc } from './formatTimestampAsUtc';
+
+export const accountVoToAccount = (accountVo: AccountVo): Account => ({
+  id: accountVo.accountId,
+  isSftpDeletionEnabled: false,
```
Fenn-CS, line 9: `isSftpDeletionEnabled: Boolean(accountVo.isSftpDeletionEnabled),`

Fenn-CS: That's if we know the name of the property in backend already, then we could avoid this needing an update anytime soon.

However, it's a small and easy change to make so passable.

Fenn-CS (approved):
Thanks for the updates Dan!

src/utils/accountVoToAccount.ts
slifty: This field doesn't exist on AccountVO yet, so I'm just setting it as false for now (we can something like what you just wrote later once it gets added to the backend!)

PR #97, opened 2023-04-18
Populate `isSftpDeletionEnabled` property

### LINKED ISSUE #96: Read sftp deletion setting
The backend now reports the user's SFTP deletion setting, so we should read that and map it to the SDK return type.

slifty:
This PR utilizes the Permanent API to populate the `Account.isSftpDeletionEnabled` property.

Resolves #96


cecilia-donnelly (approved):
:tada:

PR #99, opened 2023-04-19
Update github actions

slifty:
This PR updates our actions to their latest versions and also sets up Dependabot to keep us up to date in future.


cecilia-donnelly (approved):
Thanks for doing this! Looks great.

PR #117, opened 2023-05-18
Separate upload from archive record creation

### LINKED ISSUE #116: Separate upload from record creation
We have some SDK client user flows where the client might upload the file to s3 in other ways; we want our SDK to easily support those flows.

The current `createRecord` sdk method actually does TWO things:

1. uploads the file
2. registers the record

I'd like to propose separating this into two separate steps for sdk clients.  This also means in future the sdk might provide other ways to upload files.

slifty:
This PR separates the "upload" step from the "archive record registration" step when creating an archive record.  It means SDK clients need to either call `uploadFile` or, in the case of internal clients, upload the file to s3 directly.

Resolves #116

### DIFF
--- a/src/sdk/uploadFile.ts
+++ b/src/sdk/uploadFile.ts
@@ -1,3 +1,4 @@
+import type { Readable } from 'stream';
 import type {
   ArchiveRecord,
   File,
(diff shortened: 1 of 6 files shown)


src/sdk/uploadFile.ts
```diff
@@ -0,0 +1,41 @@
...
+import {
+  executePresignedPost,
+} from '../utils';
+
+export const uploadFile = async (
+  clientConfiguration: ClientConfiguration,
+  fileData: Buffer | Readable,
+  file: Pick<File, 'contentType' | 'size'>,
+  item: Pick<ArchiveRecord, 'displayName' | 'fileSystemCompatibleName'>,
+  parentFolder: Pick<Folder, 'id'>,
```
slifty, line 20: This doesn't seem directly relevant to uploadFile, but the permanent API itself requires it so we need to carry that requirement forward.  It does seem that there would be room for refactoring there some day (same with `item`)

liam-lloyd (approved):

PR #119, opened 2023-05-18
Move to node16 tsconfig

slifty:
This PR updates our tsconfig settings to reflect Node 16 instead of Node 14 (we are using node 16 on this project)


slifty:
Whoops I forgot we're targeting node 14+ my mistake!

PR #269, opened 2024-02-01
Add createShareLink function

liam-lloyd:
We want to expose our /share/generateShareLink endpoint through this SDK. To that end, this commit adds a new createShareLink function that hits that endpoint.


meisekimiu (commented):
I don't see this change exposing the functionality through the actual `sdk` path, which is what is ultimately publicly exported by the `permanent-sdk`. Are we waiting for the full share link features to be implemented before exposing it?

I think ideally we would have a separate type called `ShareLink` or something that would be the "cleaned up" version of the internal `ShareLinkVo` that we actually give to the end user.

liam-lloyd:
> I don't see this change exposing the functionality through the actual `sdk` path, which is what is ultimately publicly exported by the `permanent-sdk`. Are we waiting for the full share link features to be implemented before exposing it?
> 
> I think ideally we would have a separate type called `ShareLink` or something that would be the "cleaned up" version of the internal `ShareLinkVo` that we actually give to the end user.

Ah, it seems I didn't fully grasp how this project worked; updates incoming.

meisekimiu (approved):

src/api/createShareLinkVo.ts
```diff
@@ -0,0 +1,32 @@
...
+  makePermanentApiCall,
+  typedJsonParse,
+} from '../utils';
+import {
+  type ClientConfiguration,
+  type ShareLinkVo,
+  isShareLinkVo,
+} from '../types';
+
+export const createShareLinkVo = async (
```
cecilia-donnelly, line 11: @pbussi this looks like the right place to start looking

PR #273, opened 2024-02-02
Per 9426 add update share link

liam-lloyd:
Compare to per-9424_add_generate_share_link to see changes unique to this PR, if that branch isn't merged yet.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -1,0 +1,75 @@
+# API Documentation
+
+## Configuration
+Each function provided by this SDK takes a `ClientConfiguration` object:
+```
+{
+  bearerToken: string;
+  baseUrl?: string;
+}
+```
+
+`bearerToken` is required and should be a JWT corresponding to a logged in Permanent account, such as that returned by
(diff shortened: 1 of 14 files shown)


cecilia-donnelly (commented):
Thanks, @liam-lloyd ! Can you add some documentation about how to call this?

Note that in our group review session Liam showed that the best way to test these is to add a new file in the package and call the new function.

docs/API.md
```diff
@@ -0,0 +1,73 @@
...
+
+#### updateShareLink
+```
+updateShareLink(
+  configuration: ClientConfiguration,
+  shareLink: ShareLink,
+): Promise<ShareLink>
+```
+
+Each non-readonly property of the share link will be updated to match `shareLink`.
```
cecilia-donnelly, line 73: Is there still that "gotcha" where creating a share link that already exists returns an unchanged link? I know we talked about it but I don't remember what the verdict was.

liam-lloyd: It does exist for the v2 endpoint; I've added a note about it

cecilia-donnelly (approved):
Thanks @liam-lloyd !

PR #275, opened 2024-02-05
Usage docs

cecilia-donnelly:
Wrote this a while ago and I think it should be okay to merge.

### DIFF
--- a/docs/login.md
+++ b/docs/login.md
@@ -1,0 +1,40 @@
+# How to log in
+
+See [the SFTP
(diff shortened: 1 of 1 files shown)


meisekimiu (approved):
The documentation seems good to me, but I wonder if this file should go in the `docs` folder since that is where I would expect documentation to go. This could just be documentation for auth as well instead of a bigger "How To" file, if it's among other documentation files in a docs folder.

cecilia-donnelly:
Definitely! Thanks @meisekimiu , totally missed the `docs` folder.

PR #308, opened 2024-04-12
Document upload

cecilia-donnelly:
Write documentation for the upload function in the SDK. Note the mention of the encoding problem - in my testing, removing the `{encoding: 'utf8'}` line in the call to `createReadStream` made it possible to upload videos. This is true even though that's the default. I don't know why this would matter but I did include it in the notes for `uploadFile`. Maybe one of you has some insight?

Initially I thought that I would be documenting multipart upload as well, but it doesn't look like the SDK actually has multipart upload (yet). We limit files to 5GB or less over SFTP.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -83,7 +83,16 @@ uploadFile(
   params: UploadFileParams,
 ): Promise<string>
 ```
-The string returned is the file's location in S3
+The string returned is the file's location in S3.
+
+Notes: `fileSystemCompatibleName` is referred to as `downloadName`
+elsewhere. It should be unique in a given folder so that folder
+contents can be downloaded without overwriting each other. On first
+creation, use the original file or folder name, knowing that the
+backend may adjust this value to make sure it is unique in the parent
+folder. Note that we have seen errors when `fileData` is explicitly
+set to use `utf8` encoding. If using `createReadStream` to create a
+`Readable` it may be best to avoid specifying an encoding.
 
 ## Records
 
@@ -151,7 +160,10 @@ CreateArchiveRecordParams {
   parentFolder: Pick<Folder, 'id'>;
 }
 ```
-s3Url will generally be the value returned by `uploadFile`.
+
+s3Url will generally be the value returned by `uploadFile`. Run
+`createArchiveRecord` to store information about uploaded files in the
+Permanent database.
 
 ```
 createArchiveRecord(
@@ -363,3 +375,4 @@ updateShareLink(
 ```
 
 Each non-readonly property of the share link will be updated to match `shareLink`.
+


docs/API.md
```diff
@@ -73,3 +73,43 @@ updateShareLink(
 ```
 
 Each non-readonly property of the share link will be updated to match `shareLink`.
+
+#### uploadFile
```
liam-lloyd, line 77: This heading level presently makes this a subsection of the Functions subsection of the Shares section. I'd either put these both in a new Files section, or put this in a Files section and `createArchiveRecord` in a Records or ArchiveRecords section

docs/API.md
```diff
@@ -73,3 +73,43 @@ updateShareLink(
...
+
+Run this to upload the file to S3.
+
+```
+uploadFile(
+  configuration: ClientConfiguration,
+  fileData: Buffer | Readable,
+  file: contentType and size of the file to upload
+  item: displayName and fileSystemCompatibleName of the record or folder
```
liam-lloyd, line 86: I think the correct values for these fields would be clearer if we notated them as objects, or used the Typescript Pick syntax and included the whole object in a Models section

docs/API.md
```diff
@@ -73,3 +73,43 @@ updateShareLink(
...
+overwriting each other. On first creation, use the original file or
+folder name, knowing that the backend may adjust this value to make
+sure it is unique in the parent folder. Note that we have seen errors
+when `fileData` is explicitly set to use `utf8` encoding. If using
+`createReadStream` to create a `Readable` it may be best to avoid
+specifying an encoding.
+
+#### createArchiveRecord
+
+Run this to store values in the Permanent database.
```
liam-lloyd, line 104: Values doesn't seem the like the right word here. Files? Records?

cecilia-donnelly: Yeah this one is weird - I just wanted to draw a distinction between storing files in S3 and information about those files in our database. I will reword!

cecilia-donnelly:
Thank you for the comments @liam-lloyd! Made some adjustments.

docs/API.md
```diff
@@ -73,3 +73,48 @@ updateShareLink(
...
+### uploadFile
+
+Run this to upload the file to S3.
+
+```
+uploadFile(
+  configuration: ClientConfiguration,
+  fileData: Buffer | Readable,
+  file: {contentType, size}, // of the file to upload
+  item: {displayName, fileSystemCompatibleName}, // of the file or folder
```
liam-lloyd, line 88: I think we should communicate the types of the subfields here.

docs/API.md
```diff
@@ -73,3 +73,48 @@ updateShareLink(
...
+  parentFolder: id of the parent folder
```
liam-lloyd, line 89: This should also have a type, and should make clear that `parentFolder` needs to be an object with an `id` field rather than a string or number equal to the ID

docs/API.md
```diff
@@ -73,3 +73,48 @@ updateShareLink(
...
+Run this to store information about uploaded files in the Permanent
+database.
+
+```
+createArchiveRecord(
+  configuration: ClientConfiguration,
+  s3Url: string (this is returned by uploadFile)
+  file: contentType and size of the file to upload
+  item: displayName and fileSystemCompatibleName of the record or folder
+  parentFolder: id of the parent folder
```
liam-lloyd, line 117: Same comments as above

cecilia-donnelly:
Thanks @liam-lloyd ! After reading #307 I see what you're going for. I prefer a more conversational approach, but it's absolutely valuable to have the exact model in the documentation! I'll rebase once 307 is merged and just add the detail from this version.

liam-lloyd (approved):

PR #464, opened 2025-02-17
Remove support for Node 18

### LINKED ISSUE #461: Bump to Node 20
Node 18 will [no longer be supported starting in April, 2025](https://nodejs.org/en/about/previous-releases).

Since this is an SDK we should support both the LTS and current version of node (that would mean node >20).

slifty:
Node 18 is about to leave LTS status so it is time for us to move along and join the modern age of Node 20 and above.

I took this opportunity to also use `node-version-file` in our workflows which means in future we can update versions in fewer places.

I should note... I don't remember exactly how our CD / releases work for this repo; if I should add a commit to bump the SDK version to 0.9.0 say the word.

Resolves #461


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## Changed
```
liam-lloyd, line 10: If we're planning to release right after merging this (which I think we should), we should probably go ahead and add a 0.9.0 section here.

liam-lloyd (approved):

CHANGELOG.md
slifty: Sounds good (I figure I'll bump the other dependabots one last time before the formal release but that's good enough)

slifty: Actually: since I'm about to add in that new feature (the "fail on create if the name exists" flag) let's hold off on a formal release until that point.

PR #500, opened 2025-04-02
Retry calls on certain http responses

### LINKED ISSUE #149: Support retry logic
As we explore options for API throttling it might be a useful quality of life to support exponential backoff logic (and, depending on throttling implementation, support for parsing the 429 status code response to detect a `Retry-After`).

Even without throttling it would be useful to support an optional retry flag when making API calls.

This library exists and looks quite robust: https://github.com/jonbern/fetch-retry

It might be a good option as opposed to implementing this directly ourselves

slifty:
This PR makes it possible for callers to request the SDK to perform exponentially decaying retries in the event of particular HTTP responses.

It also updates our lint rules (we were using `all` which is actually not recommended)

Resolves #149

### DIFF
--- a/src/utils/getFetch.ts
+++ b/src/utils/getFetch.ts
@@ -1,0 +1,11 @@
+import originalFetch from 'node-fetch';
+import fetchRetry from 'fetch-retry';
+import type { ClientConfiguration } from '../types/sdk/ClientConfiguration';
+
+const getFetch = (clientConfiguration: ClientConfiguration) => fetchRetry(originalFetch, {
+  retries: 3,
+  retryDelay: (attempt) => (2 ** attempt) * 1000,
+  retryOn: clientConfiguration.retryOn ?? [],
+});
+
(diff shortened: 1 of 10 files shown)


src/utils/getFetch.ts
```diff
@@ -0,0 +1,11 @@
+import originalFetch from 'node-fetch';
+import fetchRetry from 'fetch-retry';
+import type { ClientConfiguration } from '../types/sdk/ClientConfiguration';
+
+const getFetch = (clientConfiguration: ClientConfiguration) => fetchRetry(originalFetch, {
+  retries: 3,
+  retryDelay: (attempt) => (2 ** attempt) * 1000,
```
slifty, line 7: I did spend a cycle trying to write a test that would check the delays / multiple retries but jest's timeout mocks broke out of my OWN time box for that effort.

liam-lloyd (approved):

PR #545, opened 2025-07-07
Enable love rules

slifty:
This PR enables our previously-disabled `love` lint ruleset / other disabled rules.

Some of these commits translated to actual code being written to properly implement various pieces of the code base where we were doing unsafe things (for example the no unsafe type casting rule exposed the lack of a proper type guard when making stela calls)

### DIFF
--- a/src/stela/getStelaFolder.ts
+++ b/src/stela/getStelaFolder.ts
@@ -1,5 +1,10 @@
-import { makeStelaApiCall } from "../utils";
-import type { ClientConfiguration, Items, StelaFolder } from "../types";
+import { makeStelaApiCall, typedJsonParse } from "../utils";
+import {
+	generateIsItems,
+	stelaFolderSchema,
+	type ClientConfiguration,
+	type StelaFolder,
+} from "../types";
 
 export const getStelaFolder = async (
 	clientConfiguration: ClientConfiguration,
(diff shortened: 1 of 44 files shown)


src/stela/getStelaFolder.ts
```diff
@@ -19,12 +19,16 @@ export const getStelaFolder = async (
...
-	);
-	if (items.length === 0) {
+	const {
+		items: [item],
+	} = typedJsonParse(responseText, generateIsItems(stelaFolderSchema));
+
+	/* eslint-disable-next-line @typescript-eslint/no-unnecessary-condition --
+	 * This is one of the documented cases where a disable is considered
+	 * necessary by the rule itself: https://typescript-eslint.io/rules/no-unnecessary-condition/#possibly-undefined-indexed-access
+	 */
```
liam-lloyd, line 29: I think this problem would be solved if we used the `strictest` base tsconfig (as we do in stela). But that would probably be a separate PR.

liam-lloyd (approved):

src/stela/getStelaFolder.ts
slifty: The only byproduct of making that change was I got to remove the disable line!

slifty: OK looks like there are a bunch of changes to be made so I'm going to do the second thing: https://github.com/PermanentOrg/permanent-sdk/issues/548

